PR #1, opened 2022-09-26
Initial setup of Review application forking from Apply

zheileman:
NOTE: to be reviewed and amended, to be fit for purpose.

I've not done any renaming etc. but thought it might be useful having this draft PR to go through it and any naming changes we have to make to be done as separate commits (we can bunch together these rename commits into a few ones, not dozens) and still have a nice "history" to be merged to master (do not squash or we will have lost this history).

In its current state tests will fail, etc. Ideally after the renaming and tweaking we will have tests/linters passing, however note we shouldn't merge this yet until we've finalised the remaining work on the CP namespace or the build pipeline will fail to push to ECR (we can also perhaps comment that part anyways until we have the proper namespace resources).


willmcb (approved):
LGTM :)

PR #13, opened 2022-10-25
Test api issues in library not app specs

timpeat:
## Description of change
Moves testing effects of datastore api responses from application specs to api client library specs. 

## Notes for reviewer
This was originally prompted by flickering specs caused by Webmock failing to reset stubs before system specs. 
I was already uncomfortable using Webmock stubs of the datastore api to determine the state of application data in the application specs, and this PR moves away from that.


willmcb (approved):
:shipit:

PR #33, opened 2022-11-14
Attempt to fix build issue

willmcb:
See:

https://github.com/sgerrand/alpine-pkg-glibc/issues/185#issuecomment-1261935191


timpeat (approved):
nice work!

PR #40, opened 2022-11-18
CRIMRE 53 User can unassign an application.

timpeat:
## Description of change
User can unassign an application.

## Link to relevant ticket
[CRIMRE-53](https://dsdmoj.atlassian.net/browse/CRIMRE-53)

Screenshot 2022-11-21 at 07 57 09

Screenshot 2022-11-21 at 07 55 36

Screenshot 2022-11-21 at 08 01 01

### DIFF
--- a/app/models/current_assignment.rb
+++ b/app/models/current_assignment.rb
@@ -1,3 +1,4 @@
+# READ MODEL
 class CurrentAssignment < ApplicationStruct
   attribute :application, Types.Instance(CrimeApplication)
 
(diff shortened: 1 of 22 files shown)


app/models/current_assignment.rb
```diff
@@ -9,23 +10,35 @@ def user_id
...
   def assigned?
-    !latest_assignment.nil?
+    !user_id.nil?
   end
 
   private
 
   def data
-    return { user_id: nil, user_name: 'Unassigned' } unless latest_assignment
+    @data ||= projection.run(Rails.application.config.event_store).fetch(:user)
```
willmcb, line 24: Would be neat to wrap this call to event store in a cleaner method with less chaining. Like a global `application_event_store` so we done have to use the full path across the app :)

app/models/current_assignment.rb
```diff
@@ -9,23 +10,35 @@ def user_id
...
+  end
 
-    latest_assignment.data
+  def projection
+    RailsEventStore::Projection
+      .from_stream("Assigning$#{application.id}")
+      .init(-> { { user: unassigned_user_data } })
+      .when(
+        Assigning::AssignedToUser,
```
willmcb, line 32: This method will get pretty long I imagine, would be good to break it up into more discreet methods that are easier to parse: 

```
  def projection
    RailsEventStore::Projection
      .get_stream
      .initialise_with_user
      .when_causes ....

```

Not sure exactly the form to take, but this at the moment looks like a case statement with the logic in the case parts rather than in separate methods.

Something to think about for a future PR in any case 👍

config/locales/en.yml
```diff
@@ -46,7 +47,8 @@ en:
       heading: All open applications
       table_headings:
         applicant_name: Applicant
-        reference: Reference 
+        reference: Ref.no.
```
willmcb, line 50: Doe we need a space here? 

Ref. No. 

Rather than 

Ref.no.

config/routes.rb
```diff
@@ -9,7 +9,7 @@
     post :assign_to_self, on: :member
   end
   
-  resources :assigned_applications, only: [:index]
+  resources :assigned_applications, only: [:index, :destroy, :create]
 
-  root 'crime_applications#index'
+  root 'assigned_applications#index'
```
willmcb, line 14: 👍

willmcb (approved):
LGTM some small non-blocking points for consideration :)

app/models/current_assignment.rb
timpeat: Agree. I imagine we'll have a mixin for event sourced classes.

app/models/current_assignment.rb
timpeat: Agree

PR #46, opened 2022-11-24
CRIMRE-68 reassign an application

timpeat:
## Description of change
Can reassign an assigned application to themselves.

## Link to relevant ticket
[CRIMRE-68](https://dsdmoj.atlassian.net/browse/CRIMRE-68)

## Notes for reviewer

Basic error for not found added. Need to bring correct pages over from apply.
app/commands in need of a tidy/refactor

### After changes:
Screenshot 2022-11-24 at 17 54 02

Confirm Page:
Screenshot 2022-11-24 at 17 54 19

Success notification
Screenshot 2022-11-24 at 17 54 30

Reasignment in history
Screenshot 2022-11-24 at 17 54 48

When reassigned to another whilst on confirm page:
Screenshot 2022-11-24 at 17 56 38

### DIFF
--- a/app/commands/assigning.rb
+++ b/app/commands/assigning.rb
@@ -1,4 +1,6 @@
 module Assigning
+  StateHasChanged = Class.new(StandardError)
+
   class Command < Dry::Struct
     attribute :crime_application_id, Types::Uuid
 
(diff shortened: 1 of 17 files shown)


app/commands/assigning.rb
```diff
@@ -1,4 +1,6 @@
 module Assigning
+  StateHasChanged = Class.new(StandardError)
```
willmcb, line 2: This syntax is interesting to me why not:

`class StateHasChanged < StandardError end` this is the more common way I've seen errors defined. We could add all the errors to a seperate file as well if we wanted one place for them

app/commands/assigning.rb
```diff
@@ -23,38 +25,112 @@ def stream_name
...
     attribute :user_name, Types::String
+    attribute :to_whom_id, Types::Uuid
+    attribute :to_whom_name, Types::String
+
+    def call
+      publish(AssignedToUser, data: to_hash)
+    end
+  end
+
+  class AssignToSelf < Assigning::Command
```
willmcb, line 36: Probably for later but it are you thinking that we might split these up into separate files - I think they may be easier to navigate. No need to do it now, thinking something like `assigning/AssignToSelf.rb` and so on

app/commands/assigning.rb
```diff
@@ -23,38 +25,112 @@ def stream_name
...
+    def validate_state
+      raise StateHasChanged if state_changed?
+    end
+
+    def state_changed?
+      state_key != current_assignment.state_key
+    end
+  end
+
+  class ReassignToUser < Assigning::Command
```
willmcb, line 85: Are reassignments the same thing as assignments? We have seperate controllers wondering if they should be seperately namespaced

app/models/application_history.rb
```diff
@@ -12,13 +20,15 @@ def load_events
     RailsEventStore::Projection
       .from_stream("Assigning$#{application.id}")
       .init(-> { [application_submitted_event] })
-      .when(
-        [Assigning::AssignedToUser, Assigning::UnassignedFromSelf],
-        ->(state, event) { state << ApplicationHistoryEvent.from_event(event) }
-      ).run(Rails.application.config.event_store).reverse
+      .when(ApplicationHistory::EVENT_TYPES,
```
willmcb, line 23: Can we chat about this

willmcb (approved):
Few non blocking comments

app/commands/assigning.rb
timpeat: Will do.

app/commands/assigning.rb
timpeat: Agree

app/models/application_history.rb
timpeat: Yes, it will possibly be gone in the next pr.

PR #54, opened 2022-12-06
Make banner and nave backgrounds full width

willmcb:
## Description of change
Makes the top banner and nav full width

Also moves 'phase banner' to the top as it looks really on otherwise and is not on the designs in any case

NOTE: the search is _less_ straight forward - would like to chat through this before I implement it. Though looking at this I think this might be mergable now as the design changes in this ticket [CRIMRE-93](https://dsdmoj.atlassian.net/browse/CRIMRE-93) remove the backgroung greyness in any case... 🤔 

## Link to relevant ticket
[CRIMRE-92](https://dsdmoj.atlassian.net/browse/CRIMRE-92)

## Notes for reviewer
- the way we use width containers probably needs to change to make the search "full width" as it will need to be outside of the `<main>` width container. Would like to chat about the best way to do this.

#### Before changes:
Screenshot 2022-12-06 at 15 43 52

#### After changes:
Screenshot 2022-12-06 at 15 43 27

#### Before changes:
Screenshot 2022-12-06 at 15 45 08

#### After changes:
TBC


timpeat (approved):
Looks good to me.

arthurashman (approved):

PR #55, opened 2022-12-07
Adds "All assigned" filter

willmcb:
## Description of change
Adds the filter to allow users to filter by all those cases that are assigned.

## Link to relevant ticket
[CRIMRE-104](https://dsdmoj.atlassian.net/browse/CRIMRE-104)

### After changes:
Screenshot 2022-12-07 at 11 55 37

- On the 'app open applications' page search for 'All assigned' and check whether the results make sense.

### DIFF
--- a/app/models/application_search.rb
+++ b/app/models/application_search.rb
@@ -30,7 +30,11 @@ class ApplicationSearch < ApplicationStruct
   def satisfied?(app, key, value)
     case key
     when :assigned_user_id
-      app.current_assignment.user_id == value
+      if value == CurrentAssignment::ALL_ASSIGNED_USER.id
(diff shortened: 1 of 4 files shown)


app/models/application_search.rb
```diff
@@ -30,7 +30,7 @@ def satisfies_all(app)
   def satisfied?(app, key, value)
     case key
     when :assigned_user_id
-      app.current_assignment.user_id == value
+      assigned_user_clause(app, value)
```
timpeat, line 33: 👍

timpeat (approved):
LGTM 💯

app/models/application_search.rb
willmcb: Yeah rubocop was complaining but I think it makes it a little easier to parse.

PR #59, opened 2022-12-13
Account for situation where there is no next application to get

arthurashman:
## Description of change

Add conditional around get next feature dependent on there being an available unassigned application for a casework to assign to themselves

## Notes for reviewer
I have made decisions unilaterally whilst awaiting response from product and design - redirecting user to their list, and showing flash message: `There are no unassigned applications to that need processing at this time, please try again later`

### Before changes:
image

### After changes:
image

Ensure there are no unassigned applications, navigate to your list page, click get next.

### DIFF
--- a/app/controllers/assigned_applications_controller.rb
+++ b/app/controllers/assigned_applications_controller.rb
@@ -18,7 +18,7 @@ class AssignedApplicationsController < ApplicationController
     )
   end
 
-  # rubocop:disable Naming/AccessorMethodName, Metrics/MethodLength
+  # rubocop:disable Naming/AccessorMethodName, Metrics/MethodLength, Metrics/AbcSize
   def get_next
     filter = ApplicationSearchFilter.new(
       {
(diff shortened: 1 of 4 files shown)


app/controllers/assigned_applications_controller.rb
```diff
@@ -18,7 +18,7 @@ def create
     )
   end
 
-  # rubocop:disable Naming/AccessorMethodName, Metrics/MethodLength
+  # rubocop:disable Naming/AccessorMethodName, Metrics/MethodLength, Metrics/AbcSize
```
willmcb, line 21: This will need a refactor once the integration with the datastore is complete is complete. No point in worrying about it now but its definitely something to come back to to break down into smaller methods or add an abstraction to reduce the method size + cyclomatic complexity of it.

willmcb (changes requested):
Just the small thing around the rubocop opt out on the test and then I think its good to go 👍

spec/system/get_next_application_spec.rb
```diff
@@ -7,4 +7,17 @@
     expect(page).to have_content('Kit Pound')
     expect(page).to have_content('This application has been assigned to you')
   end
+
+  # rubocop:disable RSpec/ExampleLength
+  it 'shows an error when there is no next application' do
+    visit '/'
+    click_on 'Get next application'
```
willmcb, line 14: You could get rid of the need for the opt out of the rubocop here like this:

```ruby
3.times do
  # the two steps
end

willmcb (approved):
Thank you :) Great job 👍

PR #61, opened 2022-12-13
CRIMRE-115 Update search

arthurashman:
## Description of change
Added subtitles in search fields, updated assigned to to caseworker, added status column, added generic spec for search fields and results.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-115

### Before changes:
image

### After changes:
image

### DIFF
--- a/app/views/application_searches/show.html.erb
+++ b/app/views/application_searches/show.html.erb
@@ -37,6 +37,9 @@
             <th scope="col" class="govuk-table__header">
               <%= t('.table_headings.caseworker') %>
             </th>
+            <th scope="col" class="govuk-table__header">
+              <%= t('.table_headings.status') %>
+            </th>
           </tr>
         </thead>
         <tbody class="govuk-table__body">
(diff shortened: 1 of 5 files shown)


config/locales/en.yml
```diff
@@ -35,8 +35,9 @@ en:
 
   labels: &LABELS
     assigned_to: Assigned to
+    application_information: Application information
+    search_criteria: Search criteria
     applicant_date_of_birth: Applicant's date of birth
-    all_caseworkers: Assigned to
```
arthurashman, line 39: This was unused as far as I could see

willmcb: Yep I think its not, probably just a leftover.

timpeat (approved):
LGTM

app/views/application_searches/show.html.erb
```diff
@@ -60,6 +63,9 @@
               <td class="govuk-table__cell">
                 <%= app.current_assignment.user_name if app.current_assignment.assigned? %>
               </td>
+              <td class="govuk-table__cell">
+                <%= app.status == 'submitted' ? 'Open' : 'Other' %>
```
willmcb, line 67: I think we could also extract this logic to the en.yml, we could have `submitted`, and the other possible status values as keys as well then just pass it the application status and it would return the right value.

See lines 90-93 in the `en.yml` to get an idea of what I'm thinking 👍

willmcb (approved):

PR #68, opened 2022-12-16
Removed success flash on get next, fixed style and layout on your list

arthurashman:

### DIFF
--- a/app/controllers/assigned_applications_controller.rb
+++ b/app/controllers/assigned_applications_controller.rb
@@ -18,7 +18,7 @@ class AssignedApplicationsController < ApplicationController
     )
   end
 
-  # rubocop:disable Naming/AccessorMethodName, Metrics/MethodLength, Metrics/AbcSize
+  # rubocop:disable Naming/AccessorMethodName, Metrics/MethodLength
   def get_next
     filter = ApplicationSearchFilter.new(
       {
@@ -35,8 +35,6 @@ class AssignedApplicationsController < ApplicationController
         user: current_user
       ).call
 
-      flash[:success] = :assigned_to_self
-
       redirect_to crime_application_path(
         id: next_app_id
       )
@@ -46,7 +44,7 @@ class AssignedApplicationsController < ApplicationController
       redirect_to assigned_applications_path
     end
   end
-  # rubocop:enable Naming/AccessorMethodName, Metrics/MethodLength, Metrics/AbcSize
+  # rubocop:enable Naming/AccessorMethodName, Metrics/MethodLength
 
   def destroy
     Assigning::UnassignFromSelf.new(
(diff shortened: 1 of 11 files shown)


app/views/assigned_applications/index.html.erb
```diff
@@ -6,44 +6,52 @@
...
+
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full govuk-!-margin-bottom-5">
-    <%= button_to(
-          'Get next application',
-          get_next_assigned_applications_path,
-          method: :post,
-          class: 'govuk-button'
-        ) %>
+    <%= form_with url: get_next_assigned_applications_path, method: :post do |f| %>
```
willmcb, line 12: Forms default to post so I don't think we need to specify it.

app/controllers/assigned_applications_controller.rb
```diff
@@ -18,7 +18,7 @@ def create
     )
   end
 
-  # rubocop:disable Naming/AccessorMethodName, Metrics/MethodLength, Metrics/AbcSize
+  # rubocop:disable Naming/AccessorMethodName, Metrics/MethodLength
```
willmcb, line 21: I think we could remove both of these opt outs:

1. Call it perhaps `next_application`
2. collapse some of these lines they aren't that long that it should matter

app/views/assigned_applications/index.html.erb
willmcb: Just thinking as well why not just keep the button with the post? there is only one element in this form so its functially in the same. I think we could have the link outside the form as it is not using the form object and then just wrap them in a row.

app/views/assigned_applications/index.html.erb
```diff
@@ -6,44 +6,52 @@
...
-        </tr>
-      </thead>
-      <tbody class="govuk-table__body">
-        <% @applications.each do |app| %>
-          <%= render 'application/crime_application', crime_application: app %>
-        <% end %>
-      </tbody>
-    </table>
+<% if @applications.any? %>
+  <div class="govuk-grid-row">
```
willmcb, line 27: Wondering if we could wrap this table in a partial - might make this view a bit easier to parse.

config/locales/en.yml
```diff
@@ -35,7 +35,7 @@ en:
     open_apps: All open applications
 
   labels: &LABELS
-    assigned_to: Assigned to
+    assigned_to: 'Assigned to:'
```
willmcb, line 38: Do we need apostrophes here? I think remove them unless we do for some reason

willmcb (commented):
Few points to consider here :)

config/locales/en.yml
arthurashman: The colon makes the yml think it's another key, so need to be in quotes. Other option apparently to do:
key: >
  Value that has: colon
  
But the quotes seems cleaner

app/controllers/assigned_applications_controller.rb
arthurashman: 👍

willmcb (approved):
LGTM :)

app/views/assigned_applications/index.html.erb
arthurashman: Done

PR #72, opened 2022-12-21
CRIMRE-118 CurrentAssignment read model.

timpeat:
Remove personal data from events.
Create Assignment Aggregate Root.
Create CurrentAssignment Read Model.

Co-authored-by: willmcb <william.mcbrien@digital.justice.gov.uk>
Co-authored-by: Arthur Ashman <arthur.ashman@digital.justice.gov.uk>


app/aggregates/assigning/commands.rb
```diff
@@ -0,0 +1,35 @@
+module Assigning
+  class AssignToUser < Assigning::Command
```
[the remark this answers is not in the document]
willmcb: Though no need to do this for this PR.

app/controllers/application_controller.rb
```diff
@@ -2,17 +2,23 @@ class ApplicationController < ActionController::Base
...
   def assignments_count
-    current_user.assigned_applications.count
+    @assignments_count ||= CurrentAssignment.where(
+      user_id: current_user_id
+    ).count
   end
 
-  def current_user
-    warden.user
+  # def current_user
```
willmcb, line 16: If this is not needed delete it 👍

app/controllers/assigned_applications_controller.rb
```diff
@@ -1,14 +1,17 @@
 class AssignedApplicationsController < ApplicationController
-  before_action :set_assigned_application, except: [:index, :create, :next_application]
-
   def index
-    @applications = current_user.assigned_applications
+    current_assignments = CurrentAssignment.where(
+      user_id: current_user_id
+    )
+
+    @applications = current_assignments.pluck(:assignment_id).map { |a_id| CrimeApplication.find(a_id) }
```
willmcb, line 7: Tiny thing `a_id` could just be `id` in the block as the fact that it is an assignment id is obvious from the `pluck` :)

app/models/application_history_item.rb
```diff
@@ -0,0 +1,48 @@
...
+        user_id: user_id,
+        user_name: User.name_for(user_id),
+        timestamp: event.timestamp,
+        event_type: event.event_type,
+        viewable_data: viewable_data_for_event(event)
+      )
+    end
+
+    def viewable_data_for_event(event)
+      case event.event_type
```
willmcb, line 35: Maybe this could be a hash look up? Might save some space - especially as we don't have a 'else' catch-all

willmcb: One thing we could do is just have this as part of the event class as `data_for_event` then just call it like so:

```ruby
def viewable_data_for_event(event)
  event.data_for_event
end
```

That way we avoid this case statement completely, and we rely on event polymophism to dictate behaviour.

app/models/application_search.rb
```diff
@@ -41,10 +41,13 @@ def satisfied?(app, key, value)
   end
 
   def assigned_user_clause(app, value)
```
willmcb, line 43: Maybe add a `?` to the end of this as it only returns booleans or things that can be evaluated to booleans.

app/models/current_assignment.rb
```diff
@@ -1,75 +1,11 @@
...
-  def assignee_from_event(event)
-    Assignee.new(
-      id: event.data.fetch(:to_whom_id, nil),
-      name: event.data.fetch(:to_whom_name, nil)
-    )
-  end
+class CurrentAssignment < ApplicationRecord
+  belongs_to :user
+
+  # def assigned_to_user?(user)
```
willmcb, line 4: Shall we remove this if its not needed?

app/models/user.rb
```diff
@@ -1,12 +1,20 @@
...
   end
 
-  # TODO: Temporary in lieu of data api decision.
-  def assigned_applications
-    CrimeApplication.all.select do |app|
-      app.current_assignment.assigned_to_user? self
+  class << self
+    # For GDPR caseworker personal data is not stored in the event stream.
+    # Rendering an applications history can require many user names to be
+    # found. This method caches those lookups.
```
willmcb, line 13: nice

config/environments/test.rb
```diff
@@ -23,7 +23,7 @@
   }
 
   # Show full error reports and disable caching.
-  config.consider_all_requests_local       = true
+  config.consider_all_requests_local       = false
```
willmcb, line 26: 👍

config/initializers/rails_event_store.rb
```diff
@@ -1,3 +1,5 @@
 Rails.configuration.to_prepare do
   Rails.configuration.event_store = RailsEventStore::Client.new
+  
+  CurrentAssignments::Configuration.new.call(Rails.configuration.event_store)
```
willmcb, line 4: 👍 nice

spec/system/reassign_an_application_confirmation_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+      assignment_id: crime_application_id
+    ).call
+
+    visit '/'
+    click_on 'All open applications'
+    click_on('Kit Pound')
+    click_on('Reassign to myself')
+  end
+
+  describe 'reloading the confirm page after a asignment has been unassigned' do
```
willmcb, line 32: needs an another 's' in assignment here 👍

willmcb (approved):
Small comments really - lets have a chat through them :)

app/aggregates/assigning/commands.rb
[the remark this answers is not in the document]
arthurashman: Agree with this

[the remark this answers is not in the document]
timpeat: Have done this. (I actually prefer them in the same file as I can compare them more easily.)

app/controllers/assigned_applications_controller.rb
timpeat: good call. will fix.

app/models/application_history_item.rb
timpeat: Agree. Will flag for a later PR.

spec/system/reassign_an_application_confirmation_spec.rb
timpeat: done

PR #73, opened 2022-12-21
Some view refactoring

willmcb:
Removed some of the looping, and use collection partials instead for things like table bodies etc.

For the changed files corresponding pages, check they render correctly


timpeat (approved):
👍

PR #74, opened 2022-12-22
Adjust spacing to match designs

arthurashman:


timpeat (approved):
Thanks for fixing! (And sorry if introduced the regression)

PR #80, opened 2023-01-06
Fix aria current page name for open applications

arthurashman:
## Description of change

Fix for aria naming to get nav highlighting working again


willmcb (approved):
:)

PR #81, opened 2023-01-06
CRIMRE-106 filter by review status

timpeat:
Co-authored-by: Arthur Ashman <arthur.ashman@digital.justice.gov.uk>

## Description of change
Adds filter by review status to search.

## Link to relevant ticket
[CRIMRE-106](https://dsdmoj.atlassian.net/browse/CRIMRE-106)

## Notes for reviewer

Review statuses are different to apply and datastore. The mapping of review status to datastore status take place in the Types.
Also uses types from LaaCrimeSchemas (these were previously duplicated).

[Datastore PR](https://github.com/ministryofjustice/laa-criminal-applications-datastore/pull/16)

### Before changes:
Screenshot 2023-01-06 at 12 08 59

### After changes:
Screenshot 2023-01-06 at 12 08 44
Screenshot 2023-01-06 at 12 08 38

[CRIMRE-106]: https://dsdmoj.atlassian.net/browse/CRIMRE-106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/lib/types.rb
+++ b/app/lib/types.rb
@@ -12,13 +12,13 @@ module Types
   # Map of review application statuses to LaaCrimeSchemas::Types:APPLICATION_STATUSES
   #
   REVIEW_APPLICATION_STATUSES = {
-    'open' => APPLICATION_STATUSES & ['submitted'],
-    'completed' => APPLICATION_STATUSES & ['completed'],
-    'sent_back' => APPLICATION_STATUSES & ['returned'],
+    'open' => [ApplicationStatus['submitted']],
+    'completed' => [], # NOTE: completed status does no yet exist in datastore/schema
+    'sent_back' => [ApplicationStatus['returned']],
     'all' => APPLICATION_STATUSES
   }.freeze
 
-  ReviewApplicationStatus = String.default('open').enum(
+  ReviewApplicationStatus = String.default('open'.freeze).enum(
     *REVIEW_APPLICATION_STATUSES.keys
   )
 
--- a/app/lib/types.rb
+++ b/app/lib/types.rb
@@ -1,64 +1,37 @@
-# TODO: share types via gem
+require 'laa_crime_schemas/types/types'
 
 module Types
-  include Dry.Types()
+  include LaaCrimeSchemas::Types
 
-  COURT_TYPES = %w[
-    crown
-    magistrates
-  ].freeze
-
-  OFFENCE_CLASSES = %w[
-    a b c d e f g h i j k
-  ].freeze
-
-  CASE_TYPES = %w[
-    summary_only
-    either_way
-    indictable
-    already_in_crown_court
-    commital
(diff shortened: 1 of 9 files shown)


app/lib/types.rb
```diff
@@ -1,64 +1,37 @@
-# TODO: share types via gem
+require 'laa_crime_schemas/types/types'
```
willmcb, line 1: 👍

app/lib/types.rb
```diff
@@ -1,64 +1,37 @@
...
   PhoneNumber = String
   Date = Date | JSON::Date
   DateTime = DateTime | JSON::DateTime
 
-  # Review only types
+  #
+  # Map of review application statuses to LaaCrimeSchemas::Types:APPLICATION_STATUSES
+  #
+  REVIEW_APPLICATION_STATUSES = {
+    'open' => APPLICATION_STATUSES & ['submitted'],
```
willmcb, line 15: This syntax is a little confusing to me as it looks like an 'and' but is really the safe navigation operator to get the values out of the APPLICATION_STATUS enum.

Do we think safe navigation is needed for this? Why not just do APPLICATION_STATUES::Submitted - or whatever the DryTypes syntax is for direct lookup of the values?

app/models/application_search_filter.rb
```diff
@@ -13,10 +14,22 @@ class ApplicationSearchFilter < ApplicationStruct
...
+  end
+
+  #
+  # Options for the application status filter
+  #
+  # Includes 'Open', 'Completed', 'Sent back to provider' or 'All applications'
+  # Values can be "open", "completed", "sent_back", "all"
+  def application_status_options
+    Types::REVIEW_APPLICATION_STATUSES.keys.map do |status|
+      [I18n.t(status, scope: 'values.application_status'), status]
```
willmcb, line 31: If we are mapping the values to keys for display purposes - why not just keep the values the same as in the schema and just display them differently in review? Why remap?

spec/system/search_applications/filter_by_status_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'rails_helper'
+
+RSpec.describe 'Search applications status filter' do
+  include_context 'when search results are returned'
+  let(:choice) { nil }
```
willmcb, line 5: I don't think this let is used in this spec?

willmcb (commented):
👍 Looks good - I would like to chat through my comments before merge :)

app/lib/types.rb
willmcb: e.g. ApplicationStatuses["foo"]

willmcb (approved):
:)

PR #88, opened 2023-01-13
Crimre 127 handle applications deleted from datastore

timpeat:
## Description of change

Your list only shows assignments if they exist in the data store.

Prior to this change, if an application was removed from the datastore whilst still assigned to a user, that user would not be able to load the "Your list" page.
 
## Link to relevant ticket
[CRIMRE-127](https://dsdmoj.atlassian.net/browse/CRIMRE-127)

## Notes for reviewer

This is a minimal solution to unblock staging should this happen again. The wider process for handing this scenario has not been determined.

This solution can result in a mismatch between the assignment counts and the number of records shown in the table. However, we do not want to delete the Assignment from review (which would correct the mismatch) because we cannot be sure that the record not being found in the datastore is intentional or just a temporary issue as a result of a bug.

[CRIMRE-127]: https://dsdmoj.atlassian.net/browse/CRIMRE-127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/assigned_applications_controller.rb
+++ b/app/controllers/assigned_applications_controller.rb
@@ -1,10 +1,9 @@
 class AssignedApplicationsController < ApplicationController
   def index
-    current_assignments = CurrentAssignment.where(
-      user_id: current_user_id
-    )
+    filter = ApplicationSearchFilter.new(assigned_status: current_user_id)
+    search = ApplicationSearch.new(filter:)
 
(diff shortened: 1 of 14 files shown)


willmcb (approved):
👍 LGTM :)

PR #94, opened 2023-01-20
Add tests for CTA work, revert lost view application format

arthurashman:
## Description of change
Added tests to ensure CTAs show on correct states, and status banners are correct per status. Also improving to ensure buttons have correct copy when assigned/unassigned/assigned to another 

Designs for status badge/assignement/dates https://www.figma.com/file/TELZW9jdNqFcrI4yZoH7BV/Crime-Review?node-id=4491%3A16813&t=hAjiuqIqCfLVzKtP-1
 
### Before changes:
image

### After changes:
image

### DIFF
--- a/app/views/crime_applications/_current_assignment.html.erb
+++ b/app/views/crime_applications/_current_assignment.html.erb
@@ -1,29 +1,32 @@
 <dl class="govuk-summary-list">
-  <dt class="govuk-summary-list__key">
-    <%= t(:assigned_to, scope: 'crime_applications.labels') %>
-  </dt>
-
-  <% if crime_application.current_assignment.nil? %>
-    <dd class="govuk-summary-list__value">
-      <%= 'Unassigned' %>
-    </dd>
-    <%= button_to t('calls_to_action.assign_to_self'),
-                  assigned_applications_path(
-                    crime_application_id: crime_application.id
-                  ),
-                  method: :post,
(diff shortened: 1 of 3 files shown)


app/views/crime_applications/_current_assignment.html.erb
```diff
@@ -1,29 +1,32 @@
...
-    <% if crime_application.current_assignment.user_id == current_user_id %>
-    <%= button_to t('calls_to_action.unassign_from_self'),
-                  assigned_application_path(crime_application),
-                  method: :delete,
-                  class: 'govuk-button govuk-button--secondary' %>
+  <dd class="govuk-summary-list__value govuk-!-width-one-half govuk-!-padding-0 govuk-body-l">
+    <div class="govuk-row">
+      <strong><%= t(:assigned_to, scope: 'crime_applications.labels') %></strong>
+      <% if crime_application.current_assignment.nil? %>
+        <%= 'Unassigned' %>
```
willmcb, line 6: Not sure this needs to be in an erb <%= as it is a string literal

willmcb (approved):
LGTM - one tiny non-blocking comment :)

app/views/crime_applications/_current_assignment.html.erb
arthurashman: Yep, meant to change that - that's how I found it

arthurashman: Wanted to put it into en.local

PR #99, opened 2023-01-25
Tidy front end

arthurashman:
## Description of change
I have ended up using this branch to resolve quite a few bits, apologies that it's not wide-reaching and not focussed

Main point of contention may be around how we consider an application's assignment after it has been returned - I have left all as is and been hacky with regard to 'Reviewed by' and conditionality of buttons etc. But I am aware of the issues leaving a closed application assigned will have on Your list application count among other issues.

### DIFF
--- a/app/views/crime_applications/_current_assignment.html.erb
+++ b/app/views/crime_applications/_current_assignment.html.erb
@@ -1,7 +1,11 @@
 <dl class="govuk-summary-list">
   <dd class="govuk-summary-list__value govuk-!-width-one-half govuk-!-padding-0 govuk-body-l">
     <div class="govuk-row">
-      <strong><%= t(:assigned_to, scope: 'crime_applications.labels') %></strong>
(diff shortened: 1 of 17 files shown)


app/views/crime_applications/_current_assignment.html.erb
```diff
@@ -12,28 +12,37 @@
       <strong><%= t(:date_received, scope: 'crime_applications.labels') %></strong>
       <%= crime_application.submitted_at.strftime('%d %B %Y') %>
     </div>
+    <% if crime_application.status == 'returned' %>
+      <div class="govuk-row">
+        <strong><%= t(:date_returned, scope: 'crime_applications.labels') %></strong>
+        <%# TODO implement returned at %>
+        24 October 2022
```
timpeat, line 19: Now available in the json of a returned application: application.return_details.returned_at

timpeat (approved):
Thank you!

PR #106, opened 2023-01-31
CRIMRE-87 Basic implementation of pagination

timpeat:
Co-authored-by: willmcb <william.mcbrien@digital.justice.gov.uk>

## Description of change

Adds pagination to search, open and closed applications, and your list.

## Link to relevant ticket
[CRIMRE-87](https://dsdmoj.atlassian.net/browse/CRIMRE-87)

[CRIMRE-87]: https://dsdmoj.atlassian.net/browse/CRIMRE-87?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -24,4 +24,18 @@ class ApplicationController < ActionController::Base
   def warden
     request.env['warden']
   end
+
+  def set_search(filter: ApplicationSearchFilter.new, default_sort_by: 'submitted_at')
+    sorting = Sorting.new(
+      permitted_params[:sorting] || { sort_by: default_sort_by }
+    )
+
+    pagination = Pagination.new(current_page: permitted_params[:page] || 1)
+
+    @search = ApplicationSearch.new(
+      filter:,
+      sorting:,
+      pagination:
+    )
+  end
 end
--- a/app/controllers/assigned_applications_controller.rb
+++ b/app/controllers/assigned_applications_controller.rb
@@ -1,11 +1,6 @@
 class AssignedApplicationsController < ApplicationController
   def index
-    sort_params = params[:sorting]&.permit(Sorting.attribute_names)
-    filter = ApplicationSearchFilter.new(assigned_status: current_user_id)
-    sorting = Sorting.new(sort_params)
-
-    @search = ApplicationSearch.new(filter:, sorting:)
-    @sorting = sorting
+    set_search(filter: ApplicationSearchFilter.new(assigned_status: current_user_id))
   end
 
   def create
@@ -23,9 +18,7 @@ class AssignedApplicationsController < ApplicationController
   end
 
   def next_application
-    filter = ApplicationSearchFilter.new(assigned_status: 'unassigned')
-    search = ApplicationSearch.new(filter:)
-    next_app_id = search.results.first&.id
+    next_app_id = GetNext.new.call
 
     if next_app_id
       Assigning::AssignToUser.new(
@@ -55,5 +48,12 @@ class AssignedApplicationsController < ApplicationController
     flash[:success] = :unassigned_from_self
 
     redirect_to assigned_applications_path
+  end
+
+  def permitted_params
+    params.permit(
+      :page,
+      sorting: Sorting.attribute_names
+    )
   end
 end
--- a/app/models/get_next.rb
+++ b/app/models/get_next.rb
@@ -1,0 +1,11 @@
+class GetNext
+  def initialize
+    filter = ApplicationSearchFilter.new(assigned_status: 'unassigned')
+    pagination = Pagination.new(limit_value: 1)
+    @search = ApplicationSearch.new(filter:, pagination:)
+  end
+
+  def call
+    @search.results.first&.id
+  end
+end
--- a/app/models/pagination.rb
+++ b/app/models/pagination.rb
@@ -1,0 +1,13 @@
+class Pagination < ApplicationStruct
+  attribute :current_page, Types::Params::Integer.optional.default(1)
+  attribute? :limit_value, Types::Params::Integer.default(50)
+  attribute? :total_pages, Types::Params::Integer
+  attribute? :total_count, Types::Params::Integer
+
+  def datastore_params
+    {
+      page: current_page,
+      per_page: limit_value
+    }
+  end
+end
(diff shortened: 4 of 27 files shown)


app/controllers/application_controller.rb
```diff
@@ -24,4 +24,18 @@ def authenticate_user!
   def warden
     request.env['warden']
   end
+
+  def set_search(filter: ApplicationSearchFilter.new, default_sort_by: 'submitted_at')
+    sorting = Sorting.new(
+      permitted_params[:sorting] || { sort_by: default_sort_by }
+    )
+
+    pagination = Pagination.new(current_page: permitted_params[:page] || 1)
```
willmcb, line 33: Perhaps we cold make the 1 here a constant in Pagination - I know its pretty obvious but some thing like:

```
Pagination::DEFAULT_PAGE
```

Would be clearer and not me we have a magic number here.

app/controllers/assigned_applications_controller.rb
```diff
@@ -1,11 +1,6 @@
 class AssignedApplicationsController < ApplicationController
   def index
-    sort_params = params[:sorting]&.permit(Sorting.attribute_names)
-    filter = ApplicationSearchFilter.new(assigned_status: current_user_id)
-    sorting = Sorting.new(sort_params)
-
-    @search = ApplicationSearch.new(filter:, sorting:)
-    @sorting = sorting
+    set_search(filter: ApplicationSearchFilter.new(assigned_status: current_user_id))
```
willmcb, line 3: So much cleaner 🧹

app/models/get_next.rb
```diff
@@ -0,0 +1,11 @@
+class GetNext
+  def initialize
```
willmcb, line 2: This is nicer makes sense to be its own thing - and will be easier to implement the real Algo 👍

app/models/pagination.rb
```diff
@@ -0,0 +1,13 @@
+class Pagination < ApplicationStruct
+  attribute :current_page, Types::Params::Integer.optional.default(1)
```
willmcb, line 2: If we have a default page here for current page, can we somehow remove the one from the `app/controllers/application_controller.rb` L33 seems like a duplication? 🤔

app/models/pagination.rb
```diff
@@ -0,0 +1,13 @@
...
+  attribute? :limit_value, Types::Params::Integer.default(50)
```
willmcb, line 3: Also think perhaps a constant could make this number clearer - what is a "limit_value" to someone looking at this for the first time. I don't think its completely obvious.

`attribute? :limit_value, Types::Params::Integer.default(APPLICATIONS_PER_PAGE)`

May clear this up a little :)

app/views/crime_applications/index.html.erb
```diff
@@ -21,7 +21,7 @@
   <table class="govuk-table app-dashboard-table">
     <thead class="govuk-table__head">
       <tr class="govuk-table__row">
-        <%= render "#{@review_status}_crime_applications_table_headers", sorting: @sorting %>
+        <%= render "#{@review_status}_crime_applications_table_headers", sorting: @search.sorting %>
```
willmcb, line 24: Where is this being set - is it from params? could someone use it to request a file they aren't supposed to have access to? 🤔

app/views/kaminari/_page.html.erb
```diff
@@ -0,0 +1,3 @@
+<li class='<%= "govuk-pagination__item #{' govuk-pagination__item--current' if page.current?} " %>'>
```
willmcb, line 1: Thinking about the path of this file do we really want it to be linked to the name "kaminari"? That to me is an unnecessary dependency - maybe we just name it:

`app/views/pagination/_page.html.erb`

That way its agnostic to the underlying implementation.

app/views/layouts/_primary_navigation.html.erb
```diff
@@ -23,16 +23,16 @@
 
           <li class="moj-primary-navigation__item">
             <%= link_to t('primary_navigation.open_apps'),
-                        crime_applications_path(status: 'open'),
+                        open_crime_applications_path,
```
willmcb, line 26: Other than the path these are identical? Could DRY it up - not sure if its worth it though unless we use it elsewhere?

config/locales/en.yml
```diff
@@ -267,3 +267,7 @@ en:
       table_headings:
         <<: *TABLE_HEADINGS
 
+  pagination:
```
willmcb, line 270: Maybe we could add an issue about looking through the en.yml and doing some spring cleaning / reorg as they have on Apply 🤔

spec/models/application_search_filter_spec.rb
```diff
@@ -49,16 +49,16 @@
...
-      it 'returns the correct datastore api search params as json' do
-        expect(to_json).to eq({
-          applicant_date_of_birth: nil, application_id_in: [],
+      it 'returns the correct datastore api search params' do
+        expect(datastore_params).to eq({
+                                         applicant_date_of_birth: nil, application_id_in: [],
             application_id_not_in: [], status: ['submitted'],
             submitted_after: nil, submitted_before: nil, search_text: nil
-        }.to_json)
+                                       })
```
willmcb, line 61: Is the spacing of the brackets here a little off? I would expect it to align with "expect"

willmcb (commented):
Good stuff! :) lets chat through my comments and get this merged in 👍

app/controllers/application_controller.rb
timpeat: 👍

app/models/pagination.rb
timpeat: 👍

app/views/crime_applications/index.html.erb
timpeat: set per action, not from params.

timpeat: Agreed to have separate views for open and closed rather than sharing index.

config/locales/en.yml
timpeat: Added an issue for doing this: https://github.com/ministryofjustice/laa-review-criminal-legal-aid/issues/107

willmcb (approved):
👍

PR #108, opened 2023-02-01
Crimre 87 set per page in views

timpeat:
## Description of change

Allows per_page to be set in the query string of any paginated page for testing and development purposes.

for example, visit:

```
/applications/open?per_page=1
```


willmcb (approved):
LGTM

PR #110, opened 2023-02-02
CRIMRE-158-superseded-as-sent-back

timpeat:
## Description of change

Treat "superseded" applications like "send_back".
Show no results when completed filter is selected.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-158

## Notes for reviewer

When a sent back application is resubmitted by a provider the original application is marked as superseded in the datastore. 

Currently (and incorrectly) review shows these applications as “superseded” and does not return them in search results when filtering with status “Sent back to provider”.

Requirements for Review: 

Superseded applications show their status as “Sent back to provider”.

Superseded applications are included in results when searching with status filter “Sent back to the provider”.

### Before changes:

Screenshot 2023-02-02 at 12 59 48
Screenshot 2023-02-02 at 12 59 40

### DIFF
--- a/app/models/application_search.rb
+++ b/app/models/application_search.rb
@@ -27,9 +27,7 @@ class ApplicationSearch
   include DatastoreApi::Traits::PaginatedResponse
 
   def datastore_search_response
-    @datastore_search_response ||= paginated_response(
-      http_client.post('/searches', datastore_params)
-    )
+    @datastore_search_response ||= paginated_response(datastore_response)
   end
 
   def datastore_params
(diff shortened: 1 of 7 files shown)


app/models/application_search.rb
```diff
@@ -39,4 +37,10 @@ def datastore_params
       sorting: @sorting.to_h
     }
   end
+
+  def datastore_response
+    http_client.post('/searches', datastore_params)
```
zheileman, line 42: Not using API client for the searches? Happy to adapt it so you can use it if needed.

timpeat: Thank you @zheileman! We're still in a bit of flux. I'll let you know as soon as we're a bit more certain about requirements!

zheileman: No prob!

willmcb (approved):
LGTM 👍

PR #116, opened 2023-02-08
Extract business day logic

timpeat:
## Description of change

Refactor daily_counts by extracting BusinessDay model.

## Notes for reviewer

Following comments on https://github.com/ministryofjustice/laa-review-criminal-legal-aid/pull/114


app/models/business_day.rb
```diff
@@ -0,0 +1,35 @@
+class BusinessDay
```
willmcb, line 1: I think this is much neater 👍 

Nice work :)

willmcb (approved):
Looks great 👍

PR #118, opened 2023-02-10
Crimre 166 update mini reports

timpeat:
## Description of change

Update open mini report to include closed applications.
Update closed mini report to show how may reports closed each day.

Lazy load reports.

## Link to relevant ticket
[CRIMRE-166](https://dsdmoj.atlassian.net/browse/CRIMRE-166)

## Notes for reviewer

Using turbo to lazy load the reports.

### Before changes:
Screenshot 2023-02-10 at 15 45 42
Screenshot 2023-02-10 at 15 45 33

### After changes:
Screenshot 2023-02-10 at 15 45 08

Screenshot 2023-02-10 at 15 44 54

[CRIMRE-166]: https://dsdmoj.atlassian.net/browse/CRIMRE-166?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/views/reporting/_daily_count.html.erb
```diff
@@ -22,7 +22,11 @@
             <% daily_count.counts.each_with_index do |count, index| %>
```
timpeat, line 22: Remove file

config/importmap.rb
```diff
@@ -1,4 +1,6 @@
 # Pin npm packages by running ./bin/importmap
+# Pin npm packages by running ./bin/importmap
```
timpeat, line 2: delete line

app/models/reporting/processed_report.rb
```diff
@@ -0,0 +1,60 @@
...
+    def initialize(day_zero: Time.zone.now.to_date)
+      @day_zero = day_zero
+      @slug = 'processed_report'
+    end
+
+    def id
+      @slug
+    end
+
+    def headers
```
willmcb, line 12: One thing I was thinking is that we could have a base class `BaseReport` and move the common methods to it:

```ruby
module Reporting
    class BaseReport
        def initialize(slug)
            @slug = slug
        end

        def id
            @slug
        end

       def headers
          # ...
       end
 
       def table
           # override
       end

      def rows
          # ...
      end
end

```

There is a little bit of duplication here that we could get rid of with this if we wanted - not sure its worth it though.

app/models/reporting/workload_report.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    def closed_applications_by_age
+      DailyCount.new(
+        filter: ApplicationSearchFilter.new(application_status: 'sent_back')
+      ).counts.map { |content| Cell.new(content) }
+    end
+
+    def days_passed
+      Array.new(number_of_rows) do |index|
+        last_row = (index + 1) == number_of_rows
```
willmcb, line 48: Could perhaps reduce this method a little like this:

```ruby
    def days_passed
      Array.new(number_of_rows) do |index|
        last_row = (index + 1) == number_of_rows

        look_up = last_row ? 'values.days_passed.last' : 'values.days_passed'

        content =  I18n.t(look_up, count: index)

        Cell.new(content, header: true)
      end
```

app/models/reporting/workload_report.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    def table
+      {
+        days_passed:,
+        open_applications_by_age:,
+        closed_applications_by_age:
+      }
+    end
+
+    def open_applications_by_age
```
willmcb, line 34: I think this would be easier to parse if we had a filter variable to pass in to the contructor:

```ruby
     filter = ApplicationSearchFilter.new(application_status: 'open')
     daily_counts = DailyCount.new(filter: filter)).counts

     daily_counts.map do |content|
        Cell.new(content)
      end
```

app/models/reporting/workload_report.rb
```diff
@@ -0,0 +1,60 @@
...
+      DailyCount.new(filter: ApplicationSearchFilter.new(application_status: 'open')).counts.map do |content|
+        Cell.new(content)
+      end
+    end
+
+    def closed_applications_by_age
+      DailyCount.new(
```
willmcb, line 41: Same here

spec/system/reporting/processed_report_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
+
+RSpec.describe 'Processed Report' do
```
willmcb, line 3: Are you planning on adding a spec that checks that we return the right values? Or does that already exist from a previous PR?

app/models/reporting/processed_report.rb
```diff
@@ -0,0 +1,60 @@
...
+    attr_reader :day_zero
+
+    def table
+      { processed_on:, applications_closed: }
+    end
+
+    #
+    # TODO: This can be calculated from review data
+    # as soon as the Review read_model is available.
+    #
```
willmcb, line 33: 👍

app/models/reporting/processed_report.rb
```diff
@@ -0,0 +1,60 @@
...
+          submitted_before: date.next_day,
+          application_status: 'sent_back'
+        )
+
+        Cell.new(ApplicationSearch.new(filter:, pagination:).total)
+      end
+    end
+
+    def dates
+      [day_zero, day_zero.prev_day, day_zero.prev_day.prev_day]
```
willmcb, line 47: can this be on seperate rows, made my eyes skip! 😆 

```ruby
[
   day_zero,
   etc...
]
```

willmcb (commented):
Nice work! I've made a few comments we can chat through on Monday, and then get this merged :)

app/models/reporting/processed_report.rb
timpeat: 👍

app/models/reporting/workload_report.rb
timpeat: 👍

spec/system/reporting/processed_report_spec.rb
timpeat: I think we'll check these on the the report unit specs, but assume well not be using the search for these values so thought I'd wait for that refactor.

willmcb (approved):
👍

PR #119, opened 2023-02-13
Prepare Review to use datastore JWT auth

zheileman:
## Description of change
Merging this will not enable JWT or change any functionality on Review, as it is not enabled yet (to enable, change `config.auth_type = :jwt`).

Once enabled, the datastore API client will make use of a little middleware introduced in the `simple-jwt-auth` gem to generate and send a JWT bearer token.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-273

## Notes for reviewer
If wanted to test locally, a corresponding PR has been raised in the datastore repo, in order to have an end to end, locally-testable demo.

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -7,7 +7,11 @@
 DATABASE_URL=postgresql://postgres@localhost/laa-review-criminal-legal-aid
 
 DEVELOPMENT_HOST=laa-review-criminal-legal-aid.test
-CRIME_APPLY_API_URL=https://laa-apply-for-criminal-legal-aid.test/api/
+
+# Local datastore endpoint
+DATASTORE_API_ROOT=http://localhost:3003
+# Local datastore API shared secret for JWT auth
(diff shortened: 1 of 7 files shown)


.env.development
```diff
@@ -7,7 +7,11 @@
 DATABASE_URL=postgresql://postgres@localhost/laa-review-criminal-legal-aid
 
 DEVELOPMENT_HOST=laa-review-criminal-legal-aid.test
-CRIME_APPLY_API_URL=https://laa-apply-for-criminal-legal-aid.test/api/
```
zheileman, line 10: I've removed this env as it seems is no longer used anywhere? Please let me know otherwise and I will revert the change!

timpeat: 👍

timpeat (approved):
Looks great! Thank you!

PR #128, opened 2023-02-16
Capture branch coverage

timpeat:
## Description of change

Capture branch coverage
Add unit test for CrimeApplication

## Notes for reviewer

This is the beginning of a wider refactor of the specs now the model interfaces are more tied down.

### DIFF
--- a/app/aggregates/reviewing/command.rb
+++ b/app/aggregates/reviewing/command.rb
@@ -19,5 +19,11 @@ module Reviewing
     def stream_name
       "Reviewing$#{application_id}"
     end
+
+    class << self
+      def call(args)
+        new(args).call
+      end
+    end
   end
 end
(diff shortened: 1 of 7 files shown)


app/aggregates/reviewing/command.rb
```diff
@@ -19,5 +19,11 @@ def repository
     def stream_name
       "Reviewing$#{application_id}"
     end
+
+    class << self
```
willmcb, line 23: Just thinking we could have a "method not implemented" stubbed call method in the base class perhaps it would signpost that commands are abstract

```
def call
  raise MethodNotImplemented
end
```

Small thing but it might be easier to undersand that the call is not defined here but in the children.

willmcb (approved):
NIce simplifies the calling here - seem my comment about the call method - what do you think?

PR #129, opened 2023-02-16
Add completed event to app history + spec

willmcb:
## Description of change
Adds completed events to an application history

## Link to relevant ticket
[CRIMRE-176](https://dsdmoj.atlassian.net/browse/CRIMRE-176)

### After changes:
Screenshot 2023-02-16 at 14 55 21

[CRIMRE-176]: https://dsdmoj.atlassian.net/browse/CRIMRE-176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -73,7 +73,7 @@ en:
     hearing_court_name: Court hearing the case
     hearing_date: Next court hearing
     home_address: Home address
-    providers_office_address: Providers office address 
+    providers_office_address: Providers office address
     correspondence_address: Correspondence address
     codefendant: Co-defendant %{number}
     offence: Offence
@@ -141,7 +141,7 @@ en:
     days_passed:
       one: 1 day
       other: "%{count} days"
-      last: "%{count} or more days" 
+      last: "%{count} or more days"
     review_status:
       submitted: Open
       open: Open
@@ -161,7 +161,7 @@ en:
       submitted: default
       open: default
       sent_back: red
-      completed: green 
+      completed: green
 
 
   calls_to_action:
@@ -173,7 +173,7 @@ en:
     search: Search
     unassign_from_self: Remove from your list
     view_all_open_applications: View all open applications
-    view_daily_count: View by days passed 
+    view_daily_count: View by days passed
     mark_complete: Mark as completed
     send_back: Send back to provider
   confirmations:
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
```diff
@@ -141,7 +141,7 @@ en:
     days_passed:
       one: 1 day
       other: "%{count} days"
-      last: "%{count} or more days" 
+      last: "%{count} or more days"
```
arthurashman, line 144: Ah, I was looking for how to do this the other day. Nice!

spec/system/viewing_an_application/history_spec.rb
```diff
@@ -121,4 +121,40 @@
...
+      stub_request(
+        :put,
+        "#{ENV.fetch('DATASTORE_API_ROOT')}/api/v2/applications/#{crime_application_id}/complete"
+      ).to_return(body: LaaCrimeSchemas.fixture(1.0).read, status: 200)
+
+      user = User.create(
+        first_name: 'Fred',
+        last_name: 'Smitheg',
+        auth_oid: '976658f9-f3d5-49ec-b0a9-485ff8b308fa',
+        email: 'Fred.Smitheg@justice.gov.uk'
```
arthurashman, line 136: Wondering if we should be using @example.com - as this could be a real email address

willmcb: I don't think it will be as the 'eg' is suffixed to me "example" or "e.g."

config/locales/en.yml
willmcb: Yep I have a vim script that removed whitespace on saving files - quite useful

arthurashman: Oh, lol I meant the other, last syntax - I didn't realise that wasn't what you changed here 😂

spec/system/viewing_an_application/history_spec.rb
arthurashman: I have just heard that it's best practice not to use real domains in emails in tests

timpeat (approved):

spec/system/viewing_an_application/history_spec.rb
willmcb: I think this iss good point - lets change this.

PR #134, opened 2023-02-17
Tidied a few presentation errors

arthurashman:
## Description of change
Fixed/tidied a few things as per designs and in discussion with Carol

Added Partner field default to no
Reverted date formats on dashboards
Reformatted offences for duplicate offences to show on their own rows one per date
Added 'Class' for offence class


timpeat (approved):
:tidy

PR #136, opened 2023-02-20
Confirm subscription

timpeat:
## Description of change

Allow api/events to confirm SNS subscription.

## Notes for reviewer

This is experimental. We're still not sure what format the AWS response body takes. Current assumption is that the root is text and the values are JSON.

Also had to work around this ruby issue: https://bugs.ruby-lang.org/issues/19158#note-10


zheileman (approved):
Cool 😄

PR #146, opened 2023-02-23
Show application_received events in history

timpeat:
## Description of change

Show application_received events in history.

## Link to relevant ticket
[CRIMRE-180](https://dsdmoj.atlassian.net/browse/CRIMRE-180)

## Notes for reviewer

These were previously fake events. 

In production environments, review will be notified via SNS that an application has been submitted. 

This notification service can be replicated in development (see datastore readme). However, to support contexts where an SNS service is not configured, review will create an application_received event when an application is first accessed by a user on review. 

The timestamp shown on the history for application_received events is the application's submittted_at and not the event's timestamp.

[CRIMRE-180]: https://dsdmoj.atlassian.net/browse/CRIMRE-180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/aggregates/reviewing/review.rb
+++ b/app/aggregates/reviewing/review.rb
@@ -4,10 +4,11 @@ module Reviewing
 
     def initialize(id)
       @id = id
-      @state = :submitted
+      @state = nil
       @return_reason = nil
       @reviewer_id = nil
       @reviewed_at = nil
+      @received_at = nil
     end
 
     attr_reader :id, :state, :return_reason, :reviewed_at, :reviewer_id
@@ -15,7 +16,7 @@ module Reviewing
     alias application_id id
 
     def receive_application(application_id:, correlation_id: nil)
-      raise AlreadyReceived unless @state == :submitted
+      raise AlreadyReceived if received?
 
       causation_id = correlation_id
 
@@ -26,6 +27,7 @@ module Reviewing
     end
 
     def send_back(application_id:, user_id:, reason:)
+      raise NotReceived unless received?
       raise AlreadySentBack if @state.equal?(:sent_back)
       raise CannotSendBackWhenCompleted if @state.equal?(:completed)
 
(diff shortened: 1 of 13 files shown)


app/aggregates/reviewing/review.rb
```diff
@@ -26,6 +27,7 @@ def receive_application(application_id:, correlation_id: nil)
     end
 
     def send_back(application_id:, user_id:, reason:)
+      raise NotReceived unless received?
       raise AlreadySentBack if @state.equal?(:sent_back)
```
willmcb, line 31: Just thinking it might be good to wrap these other guard clause conditionals in methods like `received?`

app/controllers/api/events_controller.rb
```diff
@@ -41,15 +41,15 @@ def handle_notification!
...
 
-    #
-    # TODO: simplify when notification message format decided.
-    #
     def application_id
+      message.dig('data', 'id')
+    end
+
+    def message
       if request.headers['x-amz-sns-rawdelivery'] == 'true'
```
willmcb, line 49: Could wrap this as well `is_aws_sqs_message?` or something not sure someone reading this will necessarily know what the header means.

app/models/application_history.rb
```diff
@@ -15,33 +16,23 @@ def items
 
   private
 
+  #
+  # TODO: Refactor as part of CRIMRE-180
+  # Assignment and Review will be combined into a single stream / aggregate.
```
willmcb, line 21: 👍

app/models/application_history_item.rb
```diff
@@ -5,21 +5,25 @@ class ApplicationHistoryItem < ApplicationStruct
...
   class << self
-    def from_event(event)
-      user_id = event.data.fetch(:user_id)
-      new(
-        user_name: User.name_for(user_id),
-        timestamp: event.timestamp,
-        event_type: event.event_type,
-        event_data: event.data
-      )
+    def from_event(event, application)
```
willmcb, line 15: Think we could break this up into a few named methods and make it a little easier to parse.

willmcb (approved):
Nice :) few small comments to consider but looks good :)

app/aggregates/reviewing/review.rb
timpeat: Handled in next pr.

PR #149, opened 2023-02-28
Crimre 153 resubmission history

arthurashman:
## Description of change
Recursively iterate through parent ids to collate all application history items
Built history for all items
Link to previous superseded versions via respective history application received items

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-153

image
image

Submit from apply
Assign to self, and return in review
Resubmit from apply
View application and application history

### DIFF
--- a/app/models/crime_application.rb
+++ b/app/models/crime_application.rb
@@ -46,14 +46,16 @@ class CrimeApplication < LaaCrimeSchemas::Structs::CrimeApplication
   def parent
     return nil if parent_id.nil?
 
-    CrimeApplication.find(parent_id)
+    @parent ||= CrimeApplication.find(parent_id)
   end
 
   def all_histories
+    return @all_histories if @all_histories
+
     histories = [history]
     histories += parent.all_histories if parent
 
-    histories
+    @all_histories = histories
   end
 
   private
--- a/app/models/crime_application.rb
+++ b/app/models/crime_application.rb
@@ -43,6 +43,19 @@ class CrimeApplication < LaaCrimeSchemas::Structs::CrimeApplication
     end
   end
 
+  def parent
+    return nil if parent_id.nil?
+
+    CrimeApplication.find(parent_id)
+  end
+
+  def all_histories
+    histories = [history]
+    histories += parent.all_histories if parent
+
+    histories
+  end
+
   private
 
   def applicant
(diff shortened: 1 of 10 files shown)


spec/spec_helper.rb
```diff
@@ -8,7 +8,7 @@
   coverage_criterion :branch
   # primary_coverage :branch
   minimum_coverage 100
-  add_group 'Views', 'app/views'
+  add_filter 'app/views'
```
arthurashman, line 11: To fix readout on coverage output on rspec

app/models/crime_application.rb
```diff
@@ -43,6 +43,19 @@ def find(id)
     end
   end
 
+  def parent
+    return nil if parent_id.nil?
+
+    CrimeApplication.find(parent_id)
```
timpeat, line 49: We should memoize this. Will do and push now.

spec/spec_helper.rb
willmcb: Is this flagging that the view isn't covered? If so should we not add tests rather than a filter?

timpeat (approved):

willmcb (approved):
Nice one - I have added a few comments to consider - please have a look at let me know if you would like to chat anything through 👍

app/models/crime_application.rb
```diff
@@ -43,6 +43,19 @@ def find(id)
...
+  end
+
+  def all_histories
+    histories = [history]
```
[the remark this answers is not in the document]
timpeat: Parent id is set by the datastore. But you make a REALLY good point.

app/views/crime_applications/history.html.erb
```diff
@@ -25,8 +25,10 @@
       </tr>
     </thead>
     <tbody class="govuk-table__body">
-      <% @crime_application.history.items.each do |item| %>
-        <%= render partial: item, layout: 'history_item', as: :item, locals: { application: @crime_application } %>
+      <% @crime_application.all_histories.each do |history| %>
+        <% history.items.each do |item| %>
+          <%= render partial: item, layout: 'history_item', as: :item, locals: { current_page_application_id: @crime_application.id, application: history.application } %>
```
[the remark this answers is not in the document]
timpeat: @willmcb this won't work as we need the application specific to the history here: 
```
application: history.application
```

spec/spec_helper.rb
timpeat: Testing ERB with simple coverage is very new and has some issues. We can re-enable once that has been fixed.

PR #153, opened 2023-03-02
Add header nav for application/user management navigation

arthurashman:
## Description of change
Add header nav for switching between application views and user management views
Manage user nav link needs linking up to actual controller/view

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-195

### Before changes:
image

### After changes:
image

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,6 +3,7 @@ class ApplicationController < ActionController::Base
 
   before_action :authenticate_user!
   helper_method :current_user_id
+  helper_method :current_user
   helper_method :assignments_count
 
   private
@@ -15,6 +16,10 @@ class ApplicationController < ActionController::Base
 
   def current_user_id
     warden.user.first.fetch('id')
+  end
+
+  def current_user
+    User.find(current_user_id)
   end
 
   def authenticate_user!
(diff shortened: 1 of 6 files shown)


app/controllers/application_controller.rb
```diff
@@ -17,6 +18,10 @@ def current_user_id
     warden.user.first.fetch('id')
   end
 
+  def current_user
```
timpeat, line 21: We can actually get the current_user name from ```warden.user.first.values_at('first_name', 'last_name')``` which would avoid a query if we wanted. (I imagine we'll need the active record model at some point soon anyway so no big deal). Perhaps memoize the current_user.

timpeat (approved):
Awesome!

app/controllers/application_controller.rb
```diff
@@ -17,6 +18,10 @@ def current_user_id
     warden.user.first.fetch('id')
   end
 
+  def current_user_name
+    @current_user_name || warden.user.first.values_at('first_name', 'last_name').join(' ')
```
zheileman, line 22: Heads up, this is not memoizing, needs to be `||=`
Unless this is the desired outcome.

arthurashman: Oops, thank you!

PR #154, opened 2023-03-02
CRIMRE 179 OIDC all dom1 users

timpeat:
## Link to relevant ticket
[CRIMRE-179](https://dsdmoj.atlassian.net/browse/CRIMRE-179)

## Notes for reviewer

Because we no longer check if a user is a member of the app, this PR adds a couple of additional validations on authentication to check the tennant_id matches and also confirm that the user is a member of the tenant (and not just a guest user).

I looked into using OIDC iframe session management for azure ad but it does not seem to be supported (on our tenant at least). I'll raise another ticket for intermittently checking the session via the server.

[CRIMRE-179]: https://dsdmoj.atlassian.net/browse/CRIMRE-179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


db/migrate/20230302140310_add_auth_subject_id_to_users.rb
```diff
@@ -0,0 +1,18 @@
+class AddAuthSubjectIdToUsers < ActiveRecord::Migration[7.0]
+  def change
+    # 
+    # Allow users to be created before authentication.
+    #
+    change_column_null :users, :auth_oid, true
+
+    # 
+    # Store and index a users subject claim (sub).
```
willmcb, line 9: As discussed please add a little bit more context to this comment 👍

willmcb (approved):
👍 thanks :)

PR #156, opened 2023-03-06
Add can_manage_others boolean to users table

arthurashman:
## Description of change
Add can_manage_others boolean to users table - not null, default false and indexed


timpeat (approved):
👍

PR #157, opened 2023-03-06
Restrict users to see nav option and access user admin only if allowed

arthurashman:
## Description of change

- Add conditionality to manage_users_controller to only allow users that can manage users to the dashboard - redirect to Your list and flash message
- Add can_manage_others attribute to Manage users table
- Conditionally hide Manage users link in header nav to those allowed to access

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-199

## Notes for reviewer
Check the link is not present, and that accessing` /admin/manage_users` directly redirects and flashes message

Then to be able to see the link, and access manage users:
- `rails c`
- `User.first.update(can_manage_others: true)`

### After changes:
image

### DIFF
--- a/app/controllers/admin/manage_users_controller.rb
+++ b/app/controllers/admin/manage_users_controller.rb
@@ -1,7 +1,13 @@
 module Admin
   class ManageUsersController < ApplicationController
     def index
-      @users = User.all
+      if current_user.can_manage_others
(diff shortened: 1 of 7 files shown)


timpeat (approved):
Nice!

PR #160, opened 2023-03-07
CRIMRE-200 make email case insensitive

timpeat:
## Description of change
User email are case insensitive when queried
User email has a case insensitive uniqueness constraint
User auth_subject_id has a uniqueness constraint

## Link to relevant ticket
https://dsdmoj.atlassian.net/jira/software/c/projects/CRIMRE/boards/960?selectedIssue=CRIMRE-200

## Notes for reviewer
The auth provider sends emails with uppercase characters. Users tend to provide emails in lowercase.  
Also, users are uniq by email, this uniqueness MUST ignore case.


arthurashman (approved):
Makes sense, looks good. Thank you!

PR #155, opened 2023-03-03
Crimre 188 add new user

willmcb:
## Description of change

Adds ability for a user to be added to the system

Will raise a further PR that deals with the PG::UniqueViolation error thats thrown so that it is rendered to the page on the form.

## Link to relevant ticket
[CRIMRE-188](https://dsdmoj.atlassian.net/browse/CRIMRE-188)
### After changes:

Form:
Screenshot 2023-03-07 at 15 35 49

Errors:
Screenshot 2023-03-07 at 15 36 43

## How to manually test the feature
Sign in as user with admin access

Add a new user by email

Validations:
- try adding an email that is not properly formatted and you should receive an error
- try adding a user without adding an email and you should also receive an error

[CRIMRE-188]: https://dsdmoj.atlassian.net/browse/CRIMRE-188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #168, opened 2023-03-09
Crimre 180 new users have to be unique

willmcb:
## Description of change
Follow up of [PR 155](https://github.com/ministryofjustice/laa-review-criminal-legal-aid/pull/155). Handles the error state when an admin tries to add a user that already exists on the system.

Also adds a convenience method that pretty prints `page.html` as a readable html document rather than a long string which is hard to parse.

### After changes:
Screenshot 2023-03-09 at 12 38 52

- try to add a user that already exists - you should see the error message as show in the screenshot.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -39,4 +39,3 @@ def print_page
   formatted_html = doc.to_xhtml(indent: 2)
   puts formatted_html
 end
-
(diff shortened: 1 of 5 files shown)


spec/spec_helper.rb
```diff
@@ -31,3 +31,11 @@
   config.shared_context_metadata_behavior = :apply_to_host_groups
   config.profile_examples = true
 end
+
+def print_page
```
timpeat, line 35: ?

timpeat (approved):
Looks great.

PR #171, opened 2023-03-10
Update copy, correct sp, remove unused I18N attributes

arthurashman:

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -113,18 +113,18 @@ en:
     "true": "Yes"
     "false": "No"
     not_provided: Not provided
+    deleted_user_name: '[deleted]'
     initial_application: Initial application
     home_address: Same as home address
     provider_address: Same as provider address
     providers_office_address: Same as provider address
     summary_only: Summary only
-    deleted_user_name: '[deleted]'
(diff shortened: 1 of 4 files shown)


timpeat (approved):
Tidy

config/locales/en.yml
```diff
@@ -132,7 +122,7 @@ en:
     either_way: Either way
     indictable: Indictable
     already_in_crown_court: Already in Crown Court
-    commital: Commital
+    committal: Committal
     appeal_to_crown_court: Appeal to Crown Court
     appeal_to_crown_court_with_changes: Appeal to Crown Court with charges
```
zheileman, line 127: Unrelated to the changes in this PR but I've noticed on Apply we have slightly different copy for some of these case types. Maybe worth checking if it also applies to Review?
Here are the differences:

```
already_in_crown_court: Trial already in crown court
committal: Committal for sentence
appeal_to_crown_court: Appeal to crown court
appeal_to_crown_court_with_changes: Appeal to crown court with changes in financial circumstances
```

arthurashman: Very helpful, thank you. I'll double check the designs

PR #173, opened 2023-03-15
Crimre 229 use standard omniauth OIDC strategy

timpeat:
## Description of change

Removes custom omniauth strategy and uses standard omniauth openid_connect strategy.

## Link to relevant ticket
[CRIMRE-208](https://dsdmoj.atlassian.net/browse/CRIMRE-208)


app/models/user_authenticate.rb
```diff
@@ -0,0 +1,44 @@
+class UserAuthenticate
+  def initialize(auth_hash)
+    @auth_subject_id = auth_hash.uid
+    @auth_expires_at = Time.zone.now + auth_hash.credentials.expires_in
+    @last_auth_at = Time.zone.now
+    @auth_info = auth_hash.info
+  end
```
mseedat-moj, line 7: Does the `auth_hash` need any fields customisation or are `credentials`, `uid` etc the standard response values?

mseedat-moj: There's an `auth_details ` further below so would be good to differentiate where `auth_hash` comes from versus `auth_details` - may just be a naming thing.

spec/models/user_authenticate_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+          email: 'test@example.com',
+          first_name: 'Jo',
+          last_name: 'TEST'
+        },
+        credentials: {
+          expires_in:
+        }
+        }
+      )
+    end
```
mseedat-moj, line 24: Formatting is a bit off

app/models/user_authenticate.rb
mseedat-moj: Is there a context in which `auth_hash` is nil or has missing `info` for example? E.g. because of some setting in AD?

mseedat-moj (commented):
Looks good, such general comments

app/models/user_authenticate.rb
timpeat: @mseedat-moj  uid is standard, and in this case comes from the verified id_token.
- good point on auth details, will clarify.
- a missing auth hash is handled at the rack layer by the warden strategy here https://github.com/ministryofjustice/laa-review-criminal-legal-aid/blob/main/lib/warden/strategies/azure_ad.rb#L10

willmcb (approved):
👍

PR #176, opened 2023-03-16
Sets turbo to false as default for forms

arthurashman:
## Description of change
Turbo has been imported to help with the loading of the mini-reports. It however then interfered with the forms in the app and prevented the loading of error messages when input was invalid.  We had to set turbo to false for each form that we wanted to work properly and show errors.

This PR sets turbo to false as default and then actively sets it to true on the mini reports (I think!)

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-201


arthurashman:
Duplicate

PR #188, opened 2023-03-27
Specify redirect_uri for azure_ad

timpeat:
Specify azure_ad return_uri

## Description of change

Having to specify the return_uri adds a configuration burden. 
It seemed not to be necessary when using Devise, however the host url does not seem to be being set consistently.
This change reverts to specifying the return_uri per deployment, and in doing so should hopefully fix authentication on staging.


willmcb (approved):
👍 LGTM

PR #191, opened 2023-03-29
Add missing escape character in dockerfile

hiboabd:
## Description of change
Adds missing escape character in dockerfile which should fix the build


willmcb (approved):
Thanks! :)

PR #198, opened 2023-03-31
CRIMRE-214 support numeric columns in reports

timpeat:
## Description of change

Supports numeric table columns as per gds.

## Link to relevant ticket
[CRIMRE-214](https://dsdmoj.atlassian.net/browse/CRIMRE-214)

## Notes for reviewer

When comparing two columns of numbers they should be right aligned according to  https://design-system.service.gov.uk/components/table/numbers/index.html
The text and column changes in the ticket are waiting to be finalised and will be part of a future PR.

### Before changes:
Screenshot 2023-03-31 at 15 07 59

### After changes:
Screenshot 2023-03-31 at 14 51 15

[CRIMRE-214]: https://dsdmoj.atlassian.net/browse/CRIMRE-214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/reporting/processed_report.rb
+++ b/app/models/reporting/processed_report.rb
@@ -16,8 +16,9 @@ module Reporting
 
     def applications_closed
       Array.new(3) do |days_ago|
-        date = day_zero - days_ago
-        Cell.new(closing_events.between(date...date.tomorrow).count, numeric: false)
+        date_from = day_zero - days_ago
+        date_to = date_from.tomorrow
+        Cell.new(closing_events.between(date_from...date_to).count, numeric: false)
       end
     end
 


app/models/reporting/table.rb
```diff
@@ -1,12 +1,22 @@
 module Reporting
   class Table
-    def initialize(table_hash)
+    # Table view object for presenting reports
+    # @param table_hash [Hash<Symbol, Array(String, Fixnum)>] table
+    #  column data keyed by column name
+    # @param numeric_column_keys [Array(Symbol)] numeric_column_keys,
+    #   if nil it assumes all but the first column are numeric.
+    def initialize(table_hash, numeric_column_keys: nil)
```
willmcb, line 8: 👍

app/models/reporting/table.rb
```diff
@@ -16,6 +26,17 @@ def rows
...
 
-    attr_reader :table_hash
+    attr_reader :table_hash, :numeric_column_keys
+
+    def numeric_column?(key)
+      numeric_column_keys.include?(key)
+    end
+
+    # All bar the first column are assumed to be numeric
+    def default_numeric_column_keys
```
willmcb, line 36: So say we want to have a non-numeric key for "email" or whatever the way to do this would be to pass a set of numeric keys but miss out the one we want to be non-numeric? If I have this wrong then let me know - but if this is the case is it not a bit opaque? Would it not be clearer to just explictly declare column types 🤔

config/kubernetes/staging/secrets.yml
willmcb, line 1: I think you need to rebase as this shouldn't have changed 👍

willmcb (changes requested):
Can we have a quick chat :)

app/models/reporting/table.rb
```diff
@@ -16,6 +26,17 @@ def rows
...
+      numeric_keys = table_hash.keys.dup
+      numeric_keys.shift
+      numeric_keys
```
mseedat-moj, line 39: Is the `.shift` here because the first key is a header/title?

mseedat-moj (approved):
Looks good to me, minor question from me re: shifting array

app/models/reporting/table.rb
timpeat: I'd look at it the other way round. If you want a numeric key, then specify which columns you want to be numeric.

app/models/reporting/table.rb
timpeat: Default behaviour is that all but first column is numeric.

Perhaps I should remove the default if confusing?

PR #199, opened 2023-03-31
[CRIMRE-238] Add split case reason when returning application

mseedat-moj:
## Description of change
Adds new 'split case' to list of return reasons

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-238

## Notes for reviewer
Text in figma has multiple versions so took one of them - likely be revised soon.

### Before changes:
N/A

### After changes:
Screenshot 2023-03-31 at 14 38 51

Go to Applications, scroll to the bottom and click 'Send back to provider'

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -21,7 +21,7 @@ en:
       timestamp: '%A %-d %b %H:%M'  # Monday 21 Nov 12:20
 
   phase_banner:
-    text: This is an MVP. Some things may not work as expected. 
+    text: This is an MVP. Some things may not work as expected.
     alpha: alpha
   flash:
     success:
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -179,6 +179,7 @@ en:
       duplicate_application: Duplicate application
       case_concluded: Case has already concluded
       provider_request: Provider request
+      split_case: The case has been split so we need justification for all offences
```
willmcb, line 182: This isn't quite right - from the designs:

> The case has been split, so we need justification for all offences

Please add the comma :)

willmcb (changes requested):
See tiny comme on the comma 👍

willmcb (approved):
👍

PR #204, opened 2023-04-06
[CRIMRE-185] Add SNS message verification

mseedat-moj:
## Description of change
Adds SNS message verification

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-185

## Notes for reviewer
~~Should the system throw a more specific error on rejection or would that be a security risk?~~
AWS SNS verification will cause system exception and therefore bubble up to Sentry etc but without responding back to the calling system - this could be an information leak (i.e. the crafted message cert is wrong).

Testing mechanism uses self-signed certs and mimics SNS message signing (assume real SNS messages arrive signed already) - is this too brittle a test?

N/A

N/A


key.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
willmcb, line 1: What is this used for?

I can see the keys used in the tests that are fixtures - but I dislike this being in the base dir for the app.

In this directory it has no context associated with it. Why is it there? What is it used for? Who knows.

Can we remove if it is not being used - or find a better place for it where the directory that it is in gives it some context so people who look at the app can understand its purpose.

app/controllers/api/events_controller.rb
```diff
@@ -73,5 +76,17 @@ def confirm_url
     def request_body
       @request_body ||= JSON.parse(request.body.read)
     end
+
+    def verify_request_authenticity
+      head :unauthorized if request_body.blank? || !message_verifier.authenticate!(request_body.to_json)
```
willmcb, line 81: Could you add the logic on the right side of the if statement to a named variable please.

```ruby
inauthentic_request = request_body.blank? || !message_verifier.authenticate(request_body.to_json)

head :unauthorised if inauthentic_request
```

willmcb (dismissed):

key.pem
mseedat-moj: This file is a mistake - didn't notice it, removed.

timpeat (approved):
Looks great. It would be good to also use the aws sns client to confirm subscription. But that could be added as part of a separate PR.

PR #208, opened 2023-04-13
CRIMRE-276-Hide offence class when not provided

MohammedHassanLatif93:
## Description of change
Hide offence class when not provided

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-276

## Notes for reviewer
Currently, if an offence has been manually input by a provider on Apply, there will not be an Offence Class set for it. This PR ensures that the offence class caption is only shown if it has been provided.

### Before changes:
image

### After changes:
image

- Send an application on Apply that has a manual inputted offence
- View application in Review to check there is no class caption under the offence name


spec/system/viewing_an_application/that_is_open_and_unassigned_spec.rb
```diff
@@ -54,6 +54,20 @@
...
+    let(:application_id) { '5aa4c689-6fb5-47ff-9567-5efe7f8ac211' }
+
+    it 'does not show the offence class' do
+      expect(page).not_to have_content('Class')
+    end
+  end
+
+  context 'with offence class provided' do
+    it 'does show the offence class' do
+      expect(page).to have_content('Class C')
```
willmcb, line 67: Same here look into using `within`

app/views/crime_applications/_offences.html.erb
```diff
@@ -19,10 +19,12 @@
         <tr class="govuk-table__row">
           <td scope="row" class="govuk-table__cell">
             <%= offence.name %>
-            <span class='govuk-caption-m'>
-              <%= t('crime_applications.values.class') %>
-              <%= offence.offence_class %>
-            </span>
+            <% if offence.offence_class.present? %>
```
willmcb, line 22: Nice 👍

willmcb (changes requested):
Few comments on the test to look at - can chat through them if you'd like :)

willmcb (approved):
👍

PR #210, opened 2023-04-14
CRIMRE-232 use custom OmniAuth strategy when local

timpeat:
- creates a custom DevAuth Omniauth strategy
- use DevAuth by default in development/test
- add an admin  user to db seed

## Link to relevant ticket

[CRIMRE-232](https://dsdmoj.atlassian.net/browse/CRIMRE-232)

## Notes for reviewer

Removes the need for AzureAd in development. 

Allows a developer to authenticate with any user in their database.

To use AzureAD in development set  DEV_AUTH_ENABLED=false

Includes a NonAuth'd user so developers can attempt to sign in with a non-authorised but authenticated user.

I've attempted to do this with minimal changes to the authentication code.

### Before changes:

Screenshot 2023-04-14 at 13 18 24

### After changes:

Screenshot 2023-04-14 at 16 55 00

[CRIMRE-232]: https://dsdmoj.atlassian.net/browse/CRIMRE-232?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,16 @@ You can also compile assets manually with `rails dartsass:build` at any time, an
 
 If you ever feel something is not right with the CSS or JS, run `rails assets:clobber` to purge the local cache.
 
+**5. Authenticating a user locally**
+
+After clicking "Sign in," you will be shown a list of all users in the local database. Select one to sign in as that user.
+
+To add users, select a user that can manage others and use the user management interface. A seed admin user is created by default (run ```bundle exec rails db:seed``` if none are listed). To ensure that the user's name is set correctly on first authorization, use the format "Firstname.Lastname@example.com."
+
+The last user in the list is "Not.Authorised@example.com." Select this user to simulate a non-authorized authenticated user.
+
+This authentication strategy can be disabled locally by setting ```DEV_AUTH_ENABLED=false```
+
 ## Local development with the temporary API gem
 
 The app utilises [puma-dev](https://github.com/puma/puma-dev) for local development.
--- a/app/lib/omni_auth/strategies/dev_auth.rb
+++ b/app/lib/omni_auth/strategies/dev_auth.rb
@@ -1,0 +1,67 @@
+module OmniAuth
+  module Strategies
+    class DevAuth
+      # IMPORTANT NOTE: This OmniAuth strategy is intended for local
+      # development purposes only.
+      #
+      # When a user clicks the "Sign in" button with this strategy,
+      # they are directed to the DevAuth form where they can select
+      # the email of an existing user to log in as. To add an admin
+      # user, run "bundle exec rails db:seed" if it has not been
+      # done already.
+      #
+      # The list of user emails includes a Non Authorised user's
+      # email (NO_AUTH_EMAIL), which developers can use to simulate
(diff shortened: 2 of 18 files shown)


app/lib/omni_auth/strategies/dev_auth.rb
```diff
@@ -0,0 +1,67 @@
+module OmniAuth
+  module Strategies
+    class DevAuth
+      # IMPORTANT NOTE: This OmniAuth strategy is intended for local
+      # development purposes only.
+      #
+      # When a user clicks the "Sign in" button with this strategy,
```
willmcb, line 7: Wondering ig this should go into the README rather than here?

app/lib/omni_auth/strategies/dev_auth.rb
```diff
@@ -0,0 +1,67 @@
...
+      # email (NO_AUTH_EMAIL), which developers can use to simulate
+      # a non-authorised authenticated user.
+      #
+      # During the callback phase, this strategy searches for a user
+      # with the given email address in the local database. If a user
+      # with a matching email address is found and is authenticated,
+      # the strategy constructs an auth hash based on the user's
+      # information in the database.
+      #
+      # In cases where the strategy is unable to find an authenticated
```
willmcb, line 23: Wondering if this could be more confusing - why have it guess? My thinking is that if we have the dev setup steps in the README for local dev then people will just do that. If they haven't, it doesn't work, which is a good thing. Happy either way though really, I can see this being more convenient in some ways.

app/lib/omni_auth/strategies/dev_auth.rb
```diff
@@ -0,0 +1,67 @@
...
+      # user with the given email address, it sets an auth_subject_id
+      # and guesses the user's first and last name based on the email
+      # address.
+
+      include OmniAuth::Strategy
+
+      NO_AUTH_EMAIL = 'Not.Authorised@example.com'.freeze
```
willmcb, line 30: Think we can add this to the README as well as a reference

willmcb (changes requested):
This looks great - I think we need to add some things to the README:

- The comment in the stategy, for the most part can go into the README where it talks about setting up local dev.
- similarly how to set the local ENVs and whant they mean should be there too. As well as the non-auth default email.

I think it would be good to have a list of steps to set up local login, but other than shortening the comment in the strategy I think the comments can remain in the code as they are also good signposts.

Happy to chat this through if you want :)

app/lib/omni_auth/strategies/dev_auth.rb
timpeat: done

README.md
```diff
@@ -41,6 +41,16 @@ You can also compile assets manually with `rails dartsass:build` at any time, an
 
 If you ever feel something is not right with the CSS or JS, run `rails assets:clobber` to purge the local cache.
 
+**5. Authenticating a user locally**
```
willmcb, line 44: Great :)

willmcb (approved):
LGTM 👍

PR #211, opened 2023-04-17
[CRIMRE-277] Add SNS Topic check

mseedat-moj:
Places canonical topic into environment presumably set in infrastructure code

## Description of change
Checks whether incoming message is a message we should care about

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-277

## Notes for reviewer
Assumes there is a single valid SNS Topic we wish to subscribe to set as an environment variable.

Assumes the environment variable is set during infrastructure setup and is fixed (i.e. does not change regularly)

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -20,3 +20,6 @@ DATASTORE_API_AUTH_SECRET=foobar
 
 # Notify API change here to test actual Notify calls in Dev
 GOVUK_NOTIFY_API_KEY=notify-api-key
+
+# MAAT SNS Subscription Topic ARN
+# MAAT_SNS_TOPIC_ARN=set_in_infrastructure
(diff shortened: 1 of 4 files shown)


timpeat (commented):
Looking good @mseedat-moj . 

Have we decided if we need to keep the topic arn is a secret or not (cc @willmcb @zheileman)? 

Be a good to configure staging before this is merged.

zheileman:
I think we discussed either whitelisting the whole URN, or just part of it. If whole URN, we need to expose it as an ENV variable by injecting it from the datastore namespace where this subscription is generated.

If part of it (in particular, the tenant/account ID - the numeric part) then we can go with a constant as these will not change (unless cloud platform move to another account, and that would be communicated way in advance as it will break I can imagine lots of things).

As messages are signed by Amazon, I don't see how these could be tampered with, as it means, an attacker who wanted to impersonate us, would need to deploy an AWS service on our account (numeric ID).
Perhaps we could check this with cloud platform folks? They will know better?

PR #222, opened 2023-04-18
Spec OmniAuth strategy configuration

timpeat:
Confirm OmniAuth configuration 

## Notes for reviewer

It is possible to change the configuration used by the OpenID Connect strategy in a way that still functions but undermines security.

This change adds a spec to make sure such changes are not made without full consideration.


willmcb (approved):
Thanks 👍

PR #230, opened 2023-04-21
CRIMRE 295 Ensure all applications are received

timpeat:
## Description of change

Receive the application if it has not yet been received.

In production, Review is notified of new applications via SNS. As such, applications should be "received" before being returned by the datastore.

This PR makes sure that receive_if_required! is called when loading from the datastore to support other environments which do not have SNS set up, and to act as a failsafe should an application be returned by the datastore before its SNS event message has been processed.

## Link to relevant ticket

https://dsdmoj.atlassian.net/browse/CRIMRE-295

## Notes for reviewer

To consistently reproduce the bug/issue, SNS must not be set up (i.e. local dev) and follow these steps:

1. Create a new application on Apply.
2. Without loading any other pages, go to "All open applications" on Review.
3. Select the new application from the list.
4. Confirm that the status shows as "Open".

Before these changes, the status on the applications page would initially appear as empty.


spec/shared_contexts/with_stubbed_search.rb
```diff
@@ -33,13 +33,6 @@
   end
 
   before do
-    stubbed_search_results.each do |result|
```
timpeat, line 36: This was masking the fact that some pages should have been received. The system specs now are configured as if working in a world without SNS and therefore failsafe by default.

zheileman (approved):
LGTM!

PR #231, opened 2023-04-21
Revert some previous links, correct event names

arthurashman:


timpeat (approved):
Looks good! 

As discussed, we can move this logic up into the model if/when we know which solution we'll develop.

PR #241, opened 2023-04-27
Use different Notify Service for production

timpeat:
## Description of change

- use non-production Notify service by default
- use production Notify service in production
- allow email intercept to be configured in all environments

## Link to relevant ticket

[CRIMRE-305](https://dsdmoj.atlassian.net/browse/CRIMRE-305)

## Notes for reviewer

We need to use two different Notify environments: "production" and "non-production". The production environment is used by the live service, while non-production environments are used for testing and development.

The Notify environment is determined by the value of the environment  variable 'GOVUK_NOTIFY_ENV'. If this variable is not set, the default environment is 'non-production'.

The Notify API key, set by "GOVUK_NOTIFY_API_KEY", must correspond to the Notify Service of the template set used by the environment.

This PR also modifies the interception logic so that interception can occur in all environments.

[CRIMRE-305]: https://dsdmoj.atlassian.net/browse/CRIMRE-305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zheileman (approved):
Nice! Perfect refactor.

Minor thing, unrelated really to this, but the docker-compose fails because the `GOVUK_NOTIFY_API_KEY` is not declared in there. Perhaps we can add just a dummy `GOVUK_NOTIFY_API_KEY: notify-api-key`.

I like that the initialiser for the notify config does a fetch without default because we want to be alerted if the environment where we are deploying does not have that env variable declared 👍

PR #252, opened 2023-05-05
[CRIMRE-277] Refactor Events with SnsService

mseedat-moj:
## Description of change
Introduces checks to incoming SNS message for validity before processing

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-277

## Notes for reviewer

1. Tried to refactor events_controller into a corresponding model (`ReceiveApplicationSnsEvent`), service (`Aws::SnsService`) and revised spec.
2. The refactor isn't as clean/obvious and perhaps overkill - for example, all of the behaviours currently in `Aws::SnsService` could be in `ReceiveApplicationSnsEvent`, however tried to make a distinction between the model and its own validity and the service which may have external checks (in this case, checking if the TopicARN is allowed)
3. ~~Decided against making the `ReceiveApplicationSnsEvent` inside a module such that it is named `SnsEvent::ReceiveApplication` to prevent ambiguity with `Reviewing::ReceiveApplication`~~ Decided to move the model into a namespace and called the model `SnsEvent::ReceiveApplicationEvent`- opinions welcome.
4. Checks TopicArn for all messages such as Notification and SubscriptionConfirmation, anything thrown at it really
5. Also checks the SubscribeURL belongs to the same AWS account (similar check to TopicARN check) - assuming this is a good idea? Not sure if the ARN could be different
6. Raw SNS messages are now disallowed, deliberately returns the weird `unavailable_for_legal_reasons` HTTP status (451) which hopefully means no Sentry logs to scan but consumers may have an easily identified log if things stop working

N/A

Easiest to download the branch and review the 3 classes:

1. app/controllers/api/events_controller.rb
2. app/models/sns_event/receive_application_event.rb
3. app/services/aws/sns_service.rb


zheileman (commented):
I think it looks great.

Perhaps we can add some logging in key places that would help us diagnose when something goes wrong or if there is some kind of abuse or malicious activity?

Specially in places where we are returning an error response, so we can inspect the logs and see what is going on.

.rubocop.yml
```diff
@@ -46,6 +46,9 @@ Layout/HashAlignment:
 RSpec/AnyInstance:
   Enabled: false
 
+Style/RescueModifier:
+  Enabled: false
```
zheileman, line 50: I think this is for the syntax `@var = do_something rescue {}` right?
Instead of a global disable, I think is best in this case to mark the block of code with:
```
# rubocop:disable Style/RescueModifier
[...]
# rubocop:enable Style/RescueModifier
```

Those inline rescues can quickly proliferate through the code base and sometimes they become a source of headaches 😬

app/models/sns_event/receive_application_event.rb
```diff
@@ -0,0 +1,87 @@
+module SnsEvent
+  class ReceiveApplicationEvent
+    attr_reader :headers, :request_body, :message, :last_http_status
+
+    def initialize(sns_message:, headers: {})
+      raise ArgumentError, 'Headers must be an Enumerable type' unless headers.respond_to?(:key?)
+
+      @headers = headers
+      @request_body = JSON.parse(sns_message) rescue {}
```
timpeat, line 9: Looking great! I wonder if better to shift these to their own memoized method, then we can be more explicit about what we're rescuing and logging as @zheileman points out?

timpeat (approved):
I've made a small comment, but this looks great. Thank you!

app/models/sns_event/receive_application_event.rb
mseedat-moj: I was being lazy init

mseedat-moj: @zheileman added some error output ^^^ catches StandardError because nested deserialising wont throw a JSON::ParserError but instead something like 'no implicit conversion of Hash into String'

zheileman (approved):

PR #254, opened 2023-05-09
Fix lograge payload current user method

zheileman:
## Description of change
Follow up to PR #253.

`current_user` is defined by Devise automagically. `current_user_id` is a private method in `ApplicationController` but the health controller is a `BareApplicationController` thus don't inherit this and other methods.

Fix pods going into `CrashLoopBackOff` as the probe fails to assert the readiness of the pod.


willmcb (approved):
LGTM 👍

PR #255, opened 2023-05-09
CRIMRE 209 deactivate dormant users

timpeat:
## Description of change
Prevent accounts that have not accessed the service for 6 months from signing in.

## Link to relevant ticket

https://dsdmoj.atlassian.net/browse/CRIMRE-209

## Notes for reviewer

This moves handling dormant accounts and expired invitation to after authentication so that useful notifications can be provided for the user.

### After changes:
Screenshot 2023-05-09 at 16 17 06
Screenshot 2023-05-09 at 16 15 32


willmcb (approved):
👍

PR #260, opened 2023-05-10
Add class not determined badge for unlisted offences

arthurashman:
## Description of change
Added class not determined badge for unlisted (manually input) offences. 
Required some moving of govuk layout to fit 

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-99

### Before changes:
image

### After changes:
image

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -153,9 +153,10 @@ en:
     appeal_to_crown_court: Appeal to crown court
     appeal_to_crown_court_with_changes: Appeal to crown court with changes in financial circumstances
     class: Class
+    class_not_determined: "Class\xA0not\xA0determined"
     to: to
     conflict_of_interest:
-      "yes": CONFLICT
+      "yes": Conflict
       "no": ''
     loss_of_liberty: Loss of liberty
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
```diff
@@ -153,6 +153,7 @@ en:
     appeal_to_crown_court: Appeal to crown court
     appeal_to_crown_court_with_changes: Appeal to crown court with changes in financial circumstances
     class: Class
+    class_not_determined: "CLASS\xA0NOT\xA0DETERMINED"
```
mseedat-moj, line 156: Is there no CSS way of preventing this badge from wrapping? I imagine there's no accessibility penalty by hard coding no-wrap chars like this?

zheileman, line 156: I think no need to be all uppercase, the css styling does that for you.
The escape sequence I presume is to try to keep it all in one line? Unsure how that may work on small screens but probably we can assume all case workers will use desktop screens 👍

zheileman (approved):

mseedat-moj (approved):
LGTM - minor comment about xA0 chars & accessibility impact if any

PR #261, opened 2023-05-11
CRIMRE-315 send access granted email.

timpeat:
## Description of change

Send an email to a user when they are granted access to the Crime Review service.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-315

## Notes for reviewer

See ticket for email placeholder text used.

### DIFF
--- a/app/forms/admin/new_user_form.rb
+++ b/app/forms/admin/new_user_form.rb
@@ -10,23 +10,18 @@ module Admin
     validates :can_manage_others, presence: false, inclusion: [true, nil]
 
     def save
-      return unless valid?
+      return false unless valid?
 
       # TODO: look at getting boolean from form directly
       # seems to be issue with DFE form builder
       boolean_can_manage_others = can_manage_others ? true : false
 
-      begin
-        user = User.new(
-          email: email,
-          can_manage_others: boolean_can_manage_others
-        )
-
-        user.save
-      rescue ActiveRecord::RecordNotUnique
-        errors.add(:email, :uniqueness)
-        false
-      end
+      User.create!(email: email, can_manage_others: boolean_can_manage_others)
+      NotifyMailer.access_granted_email(email).deliver_now
+      true
+    rescue ActiveRecord::RecordNotUnique
+      errors.add(:email, :uniqueness)
+      false
     end
   end
 end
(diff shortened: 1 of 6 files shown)


app/forms/admin/new_user_form.rb
```diff
@@ -10,23 +10,18 @@ class NewUserForm
...
-        user.save
-      rescue ActiveRecord::RecordNotUnique
-        errors.add(:email, :uniqueness)
-        false
-      end
+      User.create!(email: email, can_manage_others: boolean_can_manage_others)
+      NotifyMailer.access_granted_email(email).deliver_now
+      true
+    rescue ActiveRecord::RecordNotUnique
+      errors.add(:email, :uniqueness)
```
willmcb, line 23: Why not use a uniqueness validation up front? 🤔 
https://guides.rubyonrails.org/active_record_validations.html#uniqueness

app/mailers/notify_mailer.rb
```diff
@@ -27,6 +27,12 @@ def application_returned_email(application_id)
     mail(to: provider_email)
   end
 
+  def access_granted_email(email)
```
willmcb, line 30: Its kind of interesting in a way now we have two ways to create emails - I kind of like the subscription based side-effect method. I know we are not creating events for use creation, but it is a nice pattern.

Perhaps there is a refactor in there at some point where we use subscribers for these kinds of things...

Not for this PR though :)

spec/system/manage_users/add_user_spec.rb
```diff
@@ -4,6 +4,9 @@
   include_context 'when logged in user is admin'
 
   before do
+    allow(NotifyMailer).to receive(:access_granted_email) {
```
willmcb, line 7: Wonder if it might be worth having a shared helper for this particular stubb? Might DRY it out a bit we could pass the template key in.

Just a thought not blocking for this PR :)

willmcb (approved):
Few non-blocking things you might want to consider 👍

app/mailers/notify_mailer.rb
timpeat: Yes, I'd thought the same. I suspect that may be required when we add the assurance logging for user management.

app/forms/admin/new_user_form.rb
timpeat: Because we're using a database constraint which is more reliable when enforcing uniqueness. We could add an additional rails validation I suppose, but I'd want that to be alongside the database unique index.

PR #262, opened 2023-05-11
Add better email messages

willmcb:
## Description of change
Adds prose descriptors of return reasons as per designs.


arthurashman (approved):
Nice! LGTM - I'd be keen to get Louise to validate these translations though in QA before we go live

timpeat (approved):

PR #265, opened 2023-05-12
Surface the overall offence class of an application

hiboabd:
## Description of change
Following the work to calculate and persist the overall offence class of an application in the datastore, this PR surfaces the overall offence class in the application overview. Where an overall offence class has not been calculated (i.e. when one of more of the offences do not have a class available), a value of undetermined is displayed.

## Link to relevant ticket
[CRIMRE-308](https://dsdmoj.atlassian.net/browse/CRIMRE-308)

### Before changes:
Screenshot 2023-05-12 at 12 28 38
Screenshot 2023-05-12 at 12 29 03

### After changes:

Screenshot 2023-05-12 at 12 26 31
Screenshot 2023-05-12 at 12 26 16

[CRIMRE-308]: https://dsdmoj.atlassian.net/browse/CRIMRE-308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spec/fixtures/files/crime_apply_data/applications/148df27d-4710-4c5b-938c-bb132eb040ca.json
```diff
@@ -50,7 +50,7 @@
     "codefendants": [],
     "offences": [
       {
-        "class": "Class C or B",
```
hiboabd, line 53: Changed this offence to a single class offence as there is another fixture that has identical offence data and changing this meant I could test the overall offence class display for an application with single offence classes that would have an overall offence class. 

Also noticed the offence class key was incorrectly `class` so amended this to `offence_class`

zheileman: Good catch.

Wondering if some (or all) of these fixtures can be replaced with gem fixtures like we do in other places already. I think these fixtures might be just left overs from the old times.
No need to do this in this PR, can be done separately of course or at another time!

zheileman (approved):
Nice 👍

timpeat (approved):
👍

spec/fixtures/files/crime_apply_data/applications/148df27d-4710-4c5b-938c-bb132eb040ca.json
hiboabd: Ah good to know, I agree they should be replaced so we are consistent in using one approach

PR #267, opened 2023-05-12
Populate superseding history item from aggregate

timpeat:
## Description of change
Populate superseding history item from aggregate.

## Notes for reviewer

This refactor removes code to allow superseding history item to work with older data. This was required for testing as part of user testing and now no longer necessary.


zheileman (approved):
👍

willmcb (approved):
👍 great

PR #268, opened 2023-05-15
Configure staging an production to log to stdout

timpeat:
## Description of change
Configure staging an production to log to stdout


zheileman (approved):
🚀

PR #271, opened 2023-05-16
CRIMRE-325 remove httpclient dependency

timpeat:
## Description of change

Remove httpclient gem.

## Link to relevant ticket
[CRIMRE-325](https://dsdmoj.atlassian.net/browse/CRIMRE-325)

## Notes for reviewer
The httpclient gem does not support a SameSite setting of 'none', which is now required to support OIDC on some browsers. 
As a result of this dependency the logs were being filled with errors from the gem. This PR removes the httpclient dependency.

### Before changes:
Screenshot 2023-05-16 at 09 33 22


config/initializers/devise.rb
```diff
@@ -42,7 +42,7 @@
     :openid_connect,
     {
       name: :azure_ad,
-      scope: [:openid, :email, :profile],
+      scope: [:openid, :email],
       response_type: :code,
```
timpeat, line 46: "profile" scope is no longer required as we do not need to access AzureAD roles.

zheileman (approved):
🎉 Great investigatory work!

PR #272, opened 2023-05-17
Add distinguishable phase banner styling

hiboabd:
## Description of change
Adds different colours and labels on the banners for different environments. Prod is blue, Staging orange and development green. This will make them more distinguishable and help avoid general confusion.

## Link to relevant ticket
[CRIMRE-257](https://dsdmoj.atlassian.net/browse/CRIMRE-257)

## Notes for reviewer
Also removes some custom styling on the header where it was taking up the full width of the screen but this doesn't match [the design](https://design-system.service.gov.uk/components/header/)

### Before changes:
Screenshot 2023-05-17 at 11 31 55

### After changes:
Screenshot 2023-05-17 at 11 14 10
Screenshot 2023-05-17 at 11 13 53
Screenshot 2023-05-17 at 11 13 28

[CRIMRE-257]: https://dsdmoj.atlassian.net/browse/CRIMRE-257?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/kubernetes/staging/config_map.yml
```diff
@@ -13,3 +13,5 @@ data:
   OMNIAUTH_AZURE_CLIENT_ID: 5d1ba28b-234a-44e1-9464-f5bec5228bfc
   OMNIAUTH_AZURE_TENANT_ID: c6874728-71e6-41fe-a9e1-2e8c36776ad8
   OMNIAUTH_AZURE_REDIRECT_URI: https://staging.review-criminal-legal-aid.service.justice.gov.uk/users/auth/azure_ad/callback
+  PHASE_BANNER_TAG: staging
```
zheileman, line 16: Can we not use the already existing `ENV_NAME` to imply the name and the colour?
I feel these 2 additional ENV vars are unnecessary.

zheileman (commented):
Maybe we don't need the additional variables.

Also, unsure how likely this is going to change, to grant such level of parametrisation and configuration.

config/kubernetes/staging/config_map.yml
hiboabd: Discussed and agreed to remove custom env variables and use existing `HostEnv` variable

app/helpers/application_helper.rb
```diff
@@ -3,6 +3,16 @@ def service_name
     t('service.name')
   end
 
+  def app_banner_colour
```
timpeat, line 6: Any reason to add a helper for banner_colour but not banner_tag or viceversa?

timpeat (approved):
Looks okay to me. Part of me not sure about having the colour in the I18n but, also like the fact it keeps the configuration for this in the same place.

app/helpers/application_helper.rb
hiboabd: The banner tag is only used in one place so doesn't really need a helper method. I also see the banner text as unique for each environment so the method would not really do much but just return the tag whereas the app banner colour is more likely to be duplicated across environments so a method would reduce duplicating the label data - does that make sense?

app/views/layouts/govuk_template.html.erb
```diff
@@ -34,7 +34,7 @@
 
   <div class="app-phase-banner-wrapper" role="status">
     <div class="govuk-width-container">
-      <%= govuk_phase_banner(tag: { text: t('phase_banner.alpha') }, text: t('phase_banner.text'), classes: ['app-phase-banner']) %>
+      <%= govuk_phase_banner(tag: { text: t("phase_banner.tag.#{HostEnv.env_name}") }, text: t('phase_banner.text'), classes: "app-banner-#{app_banner_colour}") %>
```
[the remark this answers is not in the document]
hiboabd: Ok might be worth adding a helper method so there is a default 😄

PR #282, opened 2023-05-22
CRIMRE-292 Add differentiators for historical applications

arthurashman:
## Description of change
For superseded applications show view:
- Grey out ref number and applicant name
- Remove copy ref link
- Add warning text and link to latest application

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-292

### Before changes:
image

### After changes:
image

Return an application and resubmit it
Go to closed applications and view the version that was sent back

### DIFF
--- a/app/assets/stylesheets/local/custom.scss
+++ b/app/assets/stylesheets/local/custom.scss
@@ -18,6 +18,6 @@ form.search {
   }
 }
 
-p.help-text, #superseded  {
+p.help-text, #superseded_style  {
   color: govuk-colour("dark-grey");
 }
--- a/app/assets/stylesheets/local/custom.scss
+++ b/app/assets/stylesheets/local/custom.scss
@@ -18,6 +18,6 @@ form.search {
   }
 }
 
-p.help-text {
+p.help-text, #superseded  {
   color: govuk-colour("dark-grey");
 }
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/local/custom.scss
```diff
@@ -18,6 +18,6 @@ form.search {
   }
 }
 
-p.help-text {
+p.help-text, #superseded_style  {
```
zheileman, line 21: This would be best to be a class instead of an ID.

It also helps to keep any app-specific classes prefixed with `app`, for instance:
`p.app-help-text, app-superseded-style { ... }`

timpeat (approved):
Neat!

app/views/crime_applications/_review_overview.html.erb
```diff
@@ -6,17 +6,36 @@
       </span>
     </div>
 
-    <p class="govuk-body-l govuk-!-margin-bottom-2">
+    <% if crime_application.superseded? %>
+      <% id = 'superseded_' %>
+    <% end %>
+
+    <p class="govuk-body-l govuk-!-margin-bottom-2" id="<%= id %>style">
```
zheileman, line 13: Perhaps consider applying the class to the outer div and target elements with CSS instead of adding the class to the markup. Just a suggestion.

zheileman (approved):
A couple of suggestions, but no blocking.

PR #281, opened 2023-05-19
Only allow user managers access to service on staging

timpeat:
## Description of change
User managers can only access the service on staging.
On all other environments they are redirect to the manage users interface

## Link to relevant ticket
[CRIMRE-323](https://dsdmoj.atlassian.net/browse/CRIMRE-323)

## Notes for reviewer

I've added the authorisation request spec which is should check all routes.

## How to manually test the feature

[CRIMRE-323]: https://dsdmoj.atlassian.net/browse/CRIMRE-323?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #283, opened 2023-05-22
CRIMRE-320 Only show active users on user manager homepage.

timpeat:
## Description of change
These changes are in prep for adding separate tabs for invitated and dormant users.

- Add sub-nav for manage users interface
- Only show active users on current manage users home page
- Introduce namespaced locales for the Managed users interface
- Add an app_text helper to allow more structured local files to be accessed

## Link to relevant ticket
[CRIMRE-320](https://dsdmoj.atlassian.net/browse/CRIMRE-320)
## Notes for reviewer

This is the first of several changes taken from: https://github.com/ministryofjustice/laa-review-criminal-legal-aid/pull/281

### Before changes:
Screenshot 2023-05-22 at 13 07 28

### After changes:
Screenshot 2023-05-22 at 13 07 10


app/helpers/app_text_helper.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  def thead_text(key, options = {})
+    named_text(:table_headings, key, options)
+  end
+
+  private
+
+  def named_text(text_type, key, options = {})
+    namespace = controller.text_namespace
```
willmcb, line 13: Will this only get used with the manage users controller? Wondering if it is worth adding a superclass method that throws a `NotImplemented` error or something, just incase this is mixed into a controller that doesn't define `text_namespace` 🤔 

If we did this are there other tables we could do this in?

app/views/admin/manage_users/index.html.erb
```diff
@@ -36,10 +52,16 @@
             <td class="govuk-table__cell"><%= t(user.can_manage_others, scope: 'values') %></td>
             <td class="govuk-table__cell">
               <% unless user.id == current_user_id %>
-                <%= link_to t('.buttons.edit'), edit_admin_manage_user_path(user, return_url: request.original_url), class: 'govuk-link' %>
-                <% unless user.deactivated? %>
-                  <%= link_to t('.buttons.deactivate'), new_admin_manage_user_deactivate_users_path(user), class: 'govuk-link' %>
-                <% end %>
+                <ul class="govuk-summary-list__actions-list">
+                  <% unless user.dormant? %>
```
willmcb, line 56: Have we thought about adding pundit or cancancan for these kind of methods - perhaps not complicated enough at this point to worry about it but something to think about.

config/locales/en/manage_users.yml
```diff
@@ -0,0 +1,15 @@
+en:
+  manage_users:
```
willmcb, line 2: Nice :)

willmcb (approved):
Looks good - tiny non-blocking comment to think about 👍

app/views/admin/manage_users/index.html.erb
timpeat: Yes, I don't think we're there yet. I have a draft PR here for restricting access on production which I hope is all we need for now: https://github.com/ministryofjustice/laa-review-criminal-legal-aid/pull/281

PR #285, opened 2023-05-22
CRIMRE-320 Increase user management pages size to 50

timpeat:
## Description of change

Increases manage user pagination to 50 and simplifies return path to page 1 on all actions. 

(New design will remove need for returning to other pages on success.)

Adds a "set_flash" method for accessing namespace flash messages.

### DIFF
--- a/app/controllers/admin/manage_users_controller.rb
+++ b/app/controllers/admin/manage_users_controller.rb
@@ -9,7 +9,8 @@ module Admin
     before_action :require_user_manager!
 
     def index
-      @users = User.all.order(first_name: :asc, last_name: :asc).page(params[:page])
+      @users = User.all.order(first_name: :asc, last_name: :asc)
+                   .page(params[:page])
     end
 
     def new
--- a/config/locales/en/manage_users.yml
+++ b/config/locales/en/manage_users.yml
@@ -17,7 +17,7 @@ en:
         title: Important
         cannot_access:
           - You do not have access to that page
-          - Contact laa-crime-apply@digital.justice.gov.uk if you think this wrong
+          - Contact laa-crime-apply@digital.justice.gov.uk if you think this is wrong
 
     table_headings:
       actions: Actions
--- a/spec/system/manage_users/viewing_spec.rb
+++ b/spec/system/manage_users/viewing_spec.rb
@@ -16,7 +16,7 @@ RSpec.describe 'Manage Users Dashboard' do
     it 'show access denied flash message' do
       expect(page).to have_notification_banner(
         text: 'You do not have access to that page',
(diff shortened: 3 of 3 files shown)


app/controllers/admin/manage_users_controller.rb
```diff
@@ -7,12 +7,9 @@ def text_namespace
     end
 
     before_action :require_user_manager!
-    before_action :set_return_url, only: [:edit, :update]
 
     def index
-      page = (index_params[:page].presence || 1)
-      per_page = Rails.configuration.x.admin.pagination_per_page
-      @users = User.all.order(first_name: :asc, last_name: :asc).page(page).per(per_page)
+      @users = User.all.order(first_name: :asc, last_name: :asc).page(params[:page])
```
willmcb, line 12: Could we make this multi-line - just a bit easier to parse :)

app/controllers/application_controller.rb
```diff
@@ -31,4 +31,28 @@ def set_search(filter: ApplicationSearchFilter.new, sorting: {})
       filter:, sorting:, pagination:
     )
   end
+
+  # Sets the full flash message based on the message key.
```
willmcb, line 35: 👍

config/locales/en/manage_users.yml
```diff
@@ -8,6 +8,17 @@ en:
...
+    flash:
+      success:
+        title: Success
+        new_user_created: Email address has been added
+        user_updated: Email address has been updated
+      important:
+        title: Important
+        cannot_access:
+          - You do not have access to that page
+          - Contact laa-crime-apply@digital.justice.gov.uk if you think this wrong
```
willmcb, line 20: I think there is a missing 'is' at the end here :)

willmcb (approved):
LGTM 👍

app/controllers/admin/manage_users_controller.rb
timpeat: Done :)

app/controllers/application_controller.rb
```diff
@@ -31,4 +31,28 @@ def set_search(filter: ApplicationSearchFilter.new, sorting: {})
...
+  # Like the AppTextHelper#named_text, it will scope locales according
+  # to the text_namespace set on the controller. If none is set, it will
+  # use root.
+  #
+  # Defaults to "success: true", pass option "success: false" to override.
+  #
+  # Any other options will be passed to I18n.translate.
+  #
+  def set_flash(message_key, options = {})
+    success = options.fetch(:success, true)
```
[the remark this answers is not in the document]
timpeat: The default is to assume success is true. If set as success: false it will return false.

PR #291, opened 2023-06-01
CRIMRE-336 Update error copy

arthurashman:
## Description of change
Update error copy
Render errors directly on page, not as flash notifications

Update routing and nav rendering to be conditional on logged in status to show more helpful messaging to logged in users but not leak and info of which pages exist to external users

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-336

## Notes for reviewer
DO NOT MERGE - conversations ongoing as to whether we want to stop using flash notifications for these errors

### Before changes:
image

### After changes:
image

etc per error
To see service errors (not get redirected to user admin) set feature flag in config/settings.yml:
```
allow_user_managers_service_access:
    local: true
```
 - Doing so will make authorisation_spec and header_nav_spec fail
 
To see all error pages, remove from routes.rb:
```
, constraints:
    lambda { |_request| !Rails.application.config.consider_all_requests_local }
    ```

### DIFF
--- a/app/views/layouts/external.html.erb
+++ b/app/views/layouts/external.html.erb
@@ -1,11 +1,21 @@
 <% content_for?(:page_title) ? yield(:page_title) : fallback_title %>
 
-<% content_for(:header) do %>
-  <%= govuk_header(service_name: t('service.name'), classes: ["app-banner-#{HostEnv.env_name}"]) %>
+<% if current_user %>
+  <% content_for(:header) do %>
+    <% render partial: 'layouts/header' %>
+  <% end %>
+
+  <% content_for(:primary_navgation) do %>
+    <%= render partial: 'layouts/primary_navigation' %>
+  <% end %>
+<% else %>
+  <% content_for(:header) do %>
+    <%= govuk_header(service_name: t('service.name'), classes: ["app-banner-#{HostEnv.env_name}"]) %>
+  <% end %>
 <% end %>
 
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

File: app/views/layouts/external.html.erb [hunk 1]
```diff
@@ -1,11 +1,21 @@
 <% content_for?(:page_title) ? yield(:page_title) : fallback_title %>
```
timpeat, line 2: I think we should try to keep it so that the external layout has no current user info.  I wonder if we need to reconsider the base controller for the errors controller.

app/views/users/sessions/new.html.erb
```diff
@@ -1,3 +1,5 @@
 <% title t('landing.page_title') %>
 
-<%= govuk_start_button(text: t('calls_to_action.sign_in'), href: user_azure_ad_omniauth_authorize_path, as_button: true) %>
+<% unless flash.first&.last&.include?("You cannot access this service") %>
```
timpeat, line 3: I think we need to re-think how we handle errors depending on users authentication state etc.

timpeat (commented):
Once design/behaviour is confirmed. I think it may be worth thinking about how we separate which errors to show and what state a users session is in.

timpeat (commented):
Once design/behaviour is confirmed. I think it may be worth thinking about how we separate which errors to show and what state a users session is in.

PR #292, opened 2023-06-02
[CHORE] Check invitation_expires_at before Time comparison

mseedat-moj:
## Description of change
Causes exception if invitation_expires_at is nil for whatever reason

## Link to relevant ticket
N/A

## Notes for reviewer
Unsure if this is a general issue or due to current local development database state


arthurashman (approved):
Makes sense to me! LMK if this is in draft bc there are more changes to be made

timpeat (commented):
This should only be an issue in development. I'd rather fix the data than deploy this code.

PR #294, opened 2023-06-02
Skip modsec xss rule

arthurashman:
## Description of change
Extension of https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/pull/367 to prevent 403 forbidden on modsec xss suspicion on second modsec rule


zheileman (approved):
This seems to work, just tested myself 👍

PR #297, opened 2023-06-06
Update error copy, add nav to internal error pages

arthurashman:
## Description of change
Updated copy for existing flash messages and error pages.
Added navigation to internal error pages so users can get somewhere useful. 

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-336

### Before changes:
image

image

image

### After changes:
image

image

image

To see service errors (not get redirected to user admin) set feature flag in config/settings.yml:
```
allow_user_managers_service_access:
    local: true
```
Doing so will make authorisation_spec and header_nav_spec fail
To see all error pages, remove from routes.rb:

```
, constraints:
    lambda { |_request| !Rails.application.config.consider_all_requests_local }
```

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -1,5 +1,5 @@
-class ErrorsController < BareApplicationController
-  layout 'external'
+class ErrorsController < ApplicationController
+  layout 'application'
   before_action :authenticate_user!, except: [:forbidden]
   helper_method :assignments_count
   def application_not_found
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -1,7 +1,7 @@
 class ErrorsController < BareApplicationController
   layout 'external'
   before_action :authenticate_user!, except: [:forbidden]
-
+  helper_method :assignments_count
   def application_not_found
     respond_with_status(:not_found)
   end
@@ -25,5 +25,11 @@ class ErrorsController < BareApplicationController
       format.html { render status: }
       format.all  { head status }
     end
+  end
+
+  def assignments_count
+    @assignments_count ||= CurrentAssignment.where(
+      user_id: current_user.id
+    ).count
   end
 end
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -1,7 +1,7 @@
 class ErrorsController < ApplicationController
   layout 'application'
   before_action :authenticate_user!, except: [:forbidden]
-  helper_method :assignments_count
+
   def application_not_found
     respond_with_status(:not_found)
   end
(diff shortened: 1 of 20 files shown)


app/controllers/errors_controller.rb
```diff
@@ -1,7 +1,7 @@
 class ErrorsController < BareApplicationController
```
timpeat, line 1: Given that this is now mainly for authenticated users, I think we should review the use of the BareApplicationController here as well as the external layout.

timpeat (commented):
Looking good. I think we should stop using the external layout for the authenticated user errors as we only have one of them now.

app/controllers/errors_controller.rb
arthurashman: Great point

config/application.rb
```diff
@@ -61,5 +61,8 @@ class Application < Rails::Application
 
     # Default page size for paging results
     config.x.admin.pagination_per_page = 50
+
+    #Onboarding email address for user contact
+    config.x.admin.onboarding_email = 'LAAapplyonboarding@justice.gov.uk'
```
willmcb, line 66: What are we thinking with this for different environments?

config/locales/en/devise.yml
```diff
@@ -6,21 +6,23 @@ en:
...
         unauthenticated:
-          - You are not authorised to view this page
-          - Access to this service is restricted if a user is not signed in
+          - Sign in to view this page
         dormant: 
-          - Your access to this service has been restricted
-          - It has been more than 6 months since you last accessed the service. Your account will need to be re-activated before you can sign in.
+          - You cannot access this service
+          - This is because you have not signed in to the service for more than 6 months.
+          - Contact <a mailto=LAAapplyonboarding@justice.gov.uk>LAAapplyonboarding@justice.gov.uk</a> to reactivate your account.
```
willmcb, line 20: Can we not use the rails config value interpolated here rather than hard coding?

config/locales/en/devise.yml
```diff
@@ -6,21 +6,23 @@ en:
...
         invitation_expired: 
-          - Your invitation has expired 
-          - Invitations to access this service automatically expire after 48 hours.
+          - You cannot access this service
+          - Your invitation to access this service has expired.
+          - "Contact <a mailto=LAAapplyonboarding@justice.gov.uk>LAAapplyonboarding@justice.gov.uk</a> for a new invitation."
```
willmcb, line 24: Same here

config/application.rb
arthurashman: Thinking we shouldn't show a live email address unless it's prod?

willmcb (approved):
👍

PR #302, opened 2023-06-08
Remove back button from root, small refactor

arthurashman:
## Description of change
Fixes small bug of back button being on Your List page when it is redirected to on logon (as it is the auth root path, not your list path)
Small refactor based on @willmcb 's previous comments


willmcb (approved):
👍

PR #306, opened 2023-06-12
Amend logging filters, post audit

willmcb:
## Description of change
PR does a little rationalisation and adds some missed filters to the filter logger. Also separates the "application" domain filters from the the more technical ones (e.g. salt, passw).

I also moved these to symbols to be consistent with apply and enable comments to be added to the filter.

Analysis of what is logged across systems is linked in the comment of the ticket as a google sheet. 

## Link to relevant ticket
[CRIMRE-227](https://dsdmoj.atlassian.net/browse/CRIMRE-227)

[CRIMRE-227]: https://dsdmoj.atlassian.net/browse/CRIMRE-227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -6,33 +6,27 @@
 Rails.application.config.filter_parameters += [
   :_key, :certificate, :client_id, :client_secret,
   :oid, :otp, :passw, :pkce, :salt, :secret, :ssn,
-  :tenant_id, :token,
+  :tenant_id, :token, :crypt,
 
   # Attributes relating to an application
   # It does partial matching (i.e. `telephone_number` is covered by `phone`)
-  :address,
-  :address_line_one,
-  :address_line_two,
-  :application_start_date,
+  :address_line,
   :city,
   :code,
   :country,
-  :crypt,
   :date_of_birth,
-  :details,
   :description,
+  :details,
   :email,
   :first_name,
   :hearing_court_name,
   :hearing_date,
   :last_name,
   :lookup_id,
-  :name,
   :nino,
   :other_names,
   :phone,
   :postcode,
   :reason,
-  :submission_date,
   :urn
 ]
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -3,39 +3,36 @@
 # Configure parameters to be filtered from the log file. Use this to limit dissemination of
 # sensitive information. See the ActiveSupport::ParameterFilter documentation for supported
 # notations and behaviors.
-Rails.application.config.filter_parameters += %i[
-  _key
-  address
-  address_line_one
-  address_line_two
-  application_start_date
-  certificate
-  client_id
-  client_secret
-  code
-  country
-  crypt
-  date_of_birth
-  email
-  first_name
-  hearing_court_name
-  hearing_date
-  last_name
-  name
-  nino
-  oid
-  other_names
-  otp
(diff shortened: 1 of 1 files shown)


config/initializers/filter_parameter_logging.rb
```diff
@@ -3,39 +3,36 @@
...
+Rails.application.config.filter_parameters += [
+  :_key, :certificate, :client_id, :client_secret,
+  :oid, :otp, :passw, :pkce, :salt, :secret, :ssn,
+  :tenant_id, :token,
+
+  # Attributes relating to an application
+  # It does partial matching (i.e. `telephone_number` is covered by `phone`)
+  :address,
+  :address_line_one,
+  :address_line_two,
```
arthurashman, line 15: Wondering if we want to align this with Apply, where it is :address_line - and will pick up :address_line_one, and :address_line_two with partial matching. :address is already picking them up, but we want that to be tighter on that as it will also be picking up :email_address etc.

config/initializers/filter_parameter_logging.rb
```diff
@@ -3,39 +3,36 @@
...
+  :application_start_date,
+  :city,
+  :code,
+  :country,
+  :crypt,
```
arthurashman, line 20: Pop this one up the top

config/initializers/filter_parameter_logging.rb
```diff
@@ -3,39 +3,36 @@
...
+  :date_of_birth,
+  :details,
+  :description,
+  :email,
+  :first_name,
+  :hearing_court_name,
+  :hearing_date,
+  :last_name,
+  :lookup_id,
```
arthurashman, line 29: Is this one more appropriate to be up the top? (I don't really know what it is)

config/initializers/filter_parameter_logging.rb
```diff
@@ -3,39 +3,36 @@
...
+  :name,
+  :nino,
+  :other_names,
+  :phone,
+  :postcode,
+  :reason,
+  :submission_date,
```
zheileman, line 36: Are `application_start_date` or `submission_date` PII details or something we care filtering out? Probably not.

config/initializers/filter_parameter_logging.rb
```diff
@@ -3,39 +3,36 @@
...
+  :date_of_birth,
+  :details,
+  :description,
+  :email,
+  :first_name,
+  :hearing_court_name,
+  :hearing_date,
+  :last_name,
+  :lookup_id,
+  :name,
```
zheileman, line 30: Name is a bit too broad, I would narrow down. For instance it would be picking up `hearing_court_name` which then is redundant (are court hearing details PII anyways?, same for `hearing_date`)

config/initializers/filter_parameter_logging.rb
timpeat: It's the id of the address lookup on OS. (And can therefore be used to find the applicant's address.)

config/initializers/filter_parameter_logging.rb
willmcb: Thanks will amend

config/initializers/filter_parameter_logging.rb
willmcb: Thanks :)

timpeat (approved):
👍

PR #310, opened 2023-06-14
[CRIMRE-361] Prevent current_user from reactivating

mseedat-moj:
## Description of change
Does not allow the current logged in user from reactivating themselves by:

- Hiding the 'Reactivate' link
- Showing an error message if current_user attempts to Reactivate by entering an appropriate URL into their browser

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-361

## Notes for reviewer
This behaviour is only applicable when current_user is deactivated by somebody else but their browser session is still open.
Introduced a new helper method `gov_link_to_unless_current_user` because `link_to_if` still outputs the link just without the anchor tag.
Not sure if premature to move the reactivation methods into a dedicated controller - perhaps overkill?

#### No reactivate link for logged in user if they are deactivated but still logged in
Screenshot 2023-06-14 at 10 11 18

#### If current_user manually tries to reactivate themselves they are bounced back to the deactivated users page
Screenshot 2023-06-14 at 10 26 42

1. Log in to 2 different browsers with 2 different users, both with 'can manage others' permission set to true. 
2. Deactivate one of the users.
3. Then in the browser with the deactivated user still logged-in, browse to manage users -> deactivated users

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -56,4 +56,12 @@ module ApplicationHelper
 
     render headers, base_params:, sort_form_method:
   end
+
+  # Rails link_to_if and variants still render the link text hence custom implementation
+  # https://govuk-components.netlify.app/helpers/link/#input-erb-class-helpers
+  def gov_link_to_unless_current_user(user:, **options)
+    return '' if current_user == user
+
+    govuk_link_to options[:text], options[:path], options[:html_options]
+  end
 end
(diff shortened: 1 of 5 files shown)


app/helpers/application_helper.rb
```diff
@@ -56,4 +56,12 @@ def table_headers(column_names, search:, base_params: {}, sort_form_method: :get
 
     render headers, base_params:, sort_form_method:
   end
+
+  # Rails link_to_if and variants still render the link text hence custom implementation
```
timpeat, line 60: 👍

timpeat (approved):
Nice 💯

PR #311, opened 2023-06-14
Set time_zone to London

timpeat:
## Description of change
Set time zone to London

## Notes for reviewer

We were displaying times in UTC.


zheileman (approved):
Nice 👍

PR #312, opened 2023-06-15
Ensure all times are TZ London before formatting

timpeat:
## Description of change

Convert all timestamps to London TZ before formatting or converting to dates.

## Notes for reviewer

UTC is used for the application and database timezones across the crime review/apply services.

Because the timestamps we get from the datastore etc are in UTC, before formatting them, or converting to dates, they must be converted to the London TZ.

This is especially important for Review where we need to know specific day in the London time zone for MI.

### Before changes:

Screenshot 2023-06-15 at 14 04 21

### After changes:

Screenshot 2023-06-15 at 14 11 23

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,4 +1,13 @@
 module ApplicationHelper
+  # Convert to the London time zone before localizing.
+  def tz_l(timestamp, options = {})
+    time_zone = options.fetch(:time_zone, 'London')
+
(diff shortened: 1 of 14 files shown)


zheileman (approved):
LGTM 👍

mseedat-moj:
Looks really good @timpeat - is there a need to enforce the fact that time must be stored in datastore as UTC? Enforceable by schema perhaps? Or is it enough to only perform the timezone conversion if the date is definitely in zulu format?

PR #315, opened 2023-06-21
Allow links in flash notifications

timpeat:
## Description of change
Allow links to appear in flash messages

## Link to relevant ticket
[CRIMRE-369](https://dsdmoj.atlassian.net/browse/CRIMRE-369)

## Notes for reviewer
This issue arose after the content for some devise flash messages was updated to include a mail link.  One alternative to this change might be to include the email as a text rather than a link.

### Before changes:
Screenshot 2023-06-21 at 08 47 10

### After changes:
Screenshot 2023-06-21 at 08 46 28

[CRIMRE-369]: https://dsdmoj.atlassian.net/browse/CRIMRE-369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/flash_notice.rb
+++ b/app/models/flash_notice.rb
@@ -1,0 +1,46 @@
+# FlashNotice view object
+#
+# Takes a flash item and converts it into a view object with attributes
+# matching those used by the govuk_notification_banner helper
+
+class FlashNotice
+  include ActionView::Helpers::SanitizeHelper
+  include ActionView::Helpers::UrlHelper
+
+  def initialize(key, message)
+    content = Array[*message]
+
+    @text = content.delete_at(0)
+    @key = key
+    @raw_details = content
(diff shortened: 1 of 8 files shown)


zheileman (approved):
This makes sense to me Tim. I think unless we move to standalone pages these notifications (and I think for user management it does not make sense to be standalone pages, and notifications is nicer), if we want to show linkable email addresses then this PR solves the problem.

For discussion perhaps if instead of the email address we link to a "Contact us" page, which is also linked in the footer of the service, as we've done on Apply, and in this contact page we show the email, maybe phone number, working hours, etc. Not sure if this is something to consider for Review.

PR #318, opened 2023-06-22
Configure Content Security Policy

timpeat:
## Description of change

Configure and enforce Configure Content Security Policy

## Link to relevant ticket
[CRIMRE-371](https://dsdmoj.atlassian.net/browse/CRIMRE-371)

## Notes for reviewer

I've tested in several browsers and think I have found most violations. It would be good to test this on a DOM1 machine as soon as it is on staging.

There is a separate ticket to notify when violations occur.

[CRIMRE-371]: https://dsdmoj.atlassian.net/browse/CRIMRE-371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mseedat-moj (approved):
LGTM

config/initializers/content_security_policy.rb
```diff
@@ -4,22 +4,24 @@
...
+    policy.img_src     :self, :https, :data
+    policy.object_src  :none
+    policy.script_src  :self, :https
+    policy.style_src   :self, :https
+    # Specify URI for violation reports
+    # policy.report_uri "/csp-violation-report-endpoint"
+  end
+
+  # Generate session nonces for permitted importmap and inline scripts
+  config.content_security_policy_nonce_generator = ->(request) { SecureRandom.base64(16) }
```
zheileman, line 20: Maybe silly question but why this instead of the offered default `request.session.id.to_s` ?

app/views/layouts/govuk_template.html.erb
```diff
@@ -4,6 +4,7 @@
 
 <head>
   <title><%= yield(:page_title) %></title>
+  <%= csp_meta_tag %>
```
zheileman, line 7: Would this be best to go in the application layout in the content_for(:head) block?
On Apply we have this block (ignore the analytics line of course)

```ruby
<% content_for(:head) do %>
  <%= csrf_meta_tags %>
  <%= csp_meta_tag %>
  <%= render partial: 'layouts/analytics' if analytics_allowed? %>
<% end %>
```

zheileman: Oh good point, I forgot that!

config/initializers/content_security_policy.rb
timpeat: Using session_id felt like a bit of a compromise in order to to support get page caching (which is unlikely to ever be an issue for us).

zheileman: But would this still allow the browser to cache resources like JS and stylesheets? Without having to fetch these in each request?

app/views/layouts/govuk_template.html.erb
timpeat: I'd quite like to have a simpler static "external" layout so perhaps something to do if that happens.

PR #322, opened 2023-06-28
Reorder if statements for logic to display ioj passport vs ioj reason

arthurashman:
## Description of change
If an application that when first submitted has IoJ not needed (no reason input), if it is found to be a split case on the courts systems then it may require returning and having an IoJ reason added. Crime Apply allows for this and overrides the 'IoJ not needed' attribute locally, but it remains in place in the data on resubmission, so a resubmitted application can have both an IoJ not needed flag AND an IoJ reason attribute. 

This PR reorders the conditional statements to check first if there is an IoJ reason and render that regardless of if the not needed flag is present (from the original submission) 

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMRE-355

### DIFF
--- a/app/views/crime_applications/_interests_of_justice.html.erb
+++ b/app/views/crime_applications/_interests_of_justice.html.erb
@@ -14,18 +14,7 @@
     </tr>
   </thead>
   <tbody class="govuk-table__body">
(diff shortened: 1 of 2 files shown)


zheileman (approved):
👍

PR #323, opened 2023-06-28
Ignore trivy vulns outside our control

arthurashman:
## Description of change

These keep coming back in prometheus alerts and we are unable to fix them as it seems to be on cloud platforms side.

As a test I installed and run Trivy locally (before adding this ignore file that is!) and it didn't detect any vulns so it must have something to do with k8s.

Replica of: https://github.com/ministryofjustice/laa-criminal-applications-datastore/pull/122


zheileman (approved):
😄

PR #327, opened 2023-07-03
Workload report limited to 9 days and younger

timpeat:
## Description of change
Reduce workload report last row limit from 10 to 9 business days.

## Link to relevant ticket
[CRIMRE-392](https://dsdmoj.atlassian.net/browse/CRIMRE-392)

## Notes for reviewer

This PR fixes a bug where the label for the last row was different to the limit set on the report. This change reduces the limit to 9 Business Days and also removes the hard coded value in the label text, to hopefully prevent such a mismatch happening again.

### DIFF
--- a/app/models/reporting/workload_report.rb
+++ b/app/models/reporting/workload_report.rb
@@ -2,8 +2,12 @@ module Reporting
   class WorkloadReport
     include Reportable
 
-    def initialize(number_of_days: 4, day_zero: Time.current, last_row_limit_in_days: 10)
-      @number_of_days = number_of_days
+    # Builds a Workload Report table with 'n' number of rows. The first row corresponds
+    # to applications that were received zero business days ago. The second row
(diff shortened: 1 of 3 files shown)


app/models/reporting/workload_report.rb
```diff
@@ -37,23 +41,23 @@ def received_applications_by_age
     # 0 days
     # 1 day
     # 2 days
-    # between 3(#number_of_days) and (#last_row_limit_in_days)
+    # between 3(#number_of_rows) and (#last_row_limit_in_days)
```
arthurashman, line 44: This probably meant to stay as number_of_days

arthurashman: I still think there's something wrong here 😂 - maybe the 3 shouldn't be there. I'm being pedantic though!

arthurashman (approved):

PR #330, opened 2023-07-04
CRIMRE-374 address error handling, pages and reporting

timpeat:
## Description of change

1. Standardise error responses to reduce information leakages:

- When not authenticated, all requests to service and admin URL redirect to sign in regardless of the existing of the resource.
- When authenticated admin, all requests to the service redirect to the user admin
- When authenticated service user, all requests to user admin render 404

2. Simplify all but application_not_found error page to reduce risk of error loops.

3. Ensure error pages do not render flash messages by default.

4.  User errors controller to render all application and middleware errors.

## Link to relevant ticket
[CRIMRE-374](https://dsdmoj.atlassian.net/browse/CRIMRE-374)

See ticket for criteria. 

To render the error pages in development for testing add the following line to config/environments/development.rb 

```
config.action_dispatch.show_exceptions = true
config.consider_all_requests_local = false
```

[CRIMRE-374]: https://dsdmoj.atlassian.net/browse/CRIMRE-374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


arthurashman:
For logged in users, are we able to give them any nav options? I worry that they'll get stuck if the only way back to the app is via the service name or editing the URL 
image

timpeat:
> For logged in users, are we able to give them any nav options? I worry that they'll get stuck if the only way back to the app is via the service name or editing the URL

When logged in as a caseworker, if users come across a missing crime application, we do display the primary navigation. However, I prefer not to include the primary navigation on other error pages due to its dependency on database requests.

arthurashman:
Sorry, Ive annoyingly updated my previous comment. I realised that the ACs said we wanted the link to all open applications which would solve my issue with navigation - but the page rendered isn't that one, so something isn't quite right

arthurashman:
It's routing to page not found, not application not found

spec/system/errors_spec.rb
```diff
@@ -1,32 +1,161 @@
...
+        expect(page).to have_content 'Access to this service is restricted'
+        expect(page).to have_http_status(:forbidden)
+      end
+
+      it 'uses the simplified errors page layout' do
+        expect(page).not_to have_css('nav.moj-primary-navigation')
+      end
+    end
+
+    context 'when visiting an admin page to manage a user' do
```
willmcb, line 82: Might be good to make this more explicit - "when visiting an admin page to manage a user as a non-manager user"

spec/system/errors_spec.rb
```diff
@@ -1,32 +1,161 @@
...
+        visit '/._darcs'
+        expect(page).to have_content 'Page not found'
+        expect(page).to have_http_status :not_found
+      end
+    end
+
+    context 'when visiting a non existent page' do
+      include_context 'with an existing application'
+
+      it 'redirects to sign in even if application does not exist' do
```
willmcb, line 109: This is great - not giving anyone any clues is the best approach

spec/system/errors_spec.rb
```diff
@@ -1,32 +1,161 @@
...
+        expect(page).to have_http_status(:not_found)
+      end
+    end
   end
 
-  it 'shows the forbidden page' do
-    visit '/forbidden'
-    expect(page).to have_content 'Access to this service is restricted'
-    expect(page).to have_css('nav.moj-primary-navigation')
+  context 'when logged in admin' do
```
willmcb, line 139: Should this be - "when logged in as an admin" or "when logged in to the admin" interface as we aren't using the admin as a user type really then the latter might be the best approach.

willmcb (approved):
Thanks looks really good - have have added a few non-blocking comments for you to consider :)

spec/system/errors_spec.rb
timpeat: This is wrapped in the context "when not authenticated"

PR #338, opened 2023-07-06
Remove spaces from nino

willmcb:
## Description of change
Removes the NINO spaces from review so that the are easier to copy for CWs etc.

## Link to relevant ticket
[CRIMRE-388](https://dsdmoj.atlassian.net/browse/CRIMRE-388)

## Notes for reviewer
I have left the formatting methof in as it:
1. removes some of the dot chain that the crime application has to go through to get to the nino
2. has some logic in it to check the nil-ness of the nino for the view.

Happy to discuss removing this if you think its not needed 👍 

### Before changes:
Screenshot 2023-07-06 at 15 30 49

### After changes:
Screenshot 2023-07-06 at 15 30 12

Look at an applications Client details section, make sure the NINO has no spaces in it.

### DIFF
--- a/app/models/crime_application.rb
+++ b/app/models/crime_application.rb
@@ -11,10 +11,7 @@ class CrimeApplication < LaaCrimeSchemas::Structs::CrimeApplication
   def formatted_applicant_nino
     return if applicant.nino.nil?
 
-    # Remove all spaces
-    formatted_nino = applicant.nino.gsub(/\s+/, '').upcase
-    [2, 5, 8, 11].each { |i| formatted_nino.insert i, ' ' } if formatted_nino.length == 9
-    formatted_nino
+    applicant.nino
   end
 
   def formatted_applicant_telephone_number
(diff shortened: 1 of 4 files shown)


timpeat (approved):
🚀

app/models/crime_application.rb
```diff
@@ -11,10 +11,7 @@ def applicant_name
   def formatted_applicant_nino
     return if applicant.nino.nil?
 
-    # Remove all spaces
-    formatted_nino = applicant.nino.gsub(/\s+/, '').upcase
-    [2, 5, 8, 11].each { |i| formatted_nino.insert i, ' ' } if formatted_nino.length == 9
-    formatted_nino
+    applicant.nino
```
arthurashman, line 14: Are we happy that Apply/the datastore removes any user-input spacing?

willmcb: Yep I checked the validation in Apply and it does

arthurashman (approved):

PR #340, opened 2023-07-07
User role cannot be changed

timpeat:
## Description of change

Once created, service users cannot become user managers and vice versa.  

## Link to relevant ticket
[CRIMRE-379](https://dsdmoj.atlassian.net/browse/CRIMRE-379)

## Notes for reviewer
The feature to change users' roles was introduced during the initial development of the user management interface. However, this addition presents certain risks and complexities, particularly in terms of comprehending a user's past privileges. We do not foresee a need for this capability in the live service.

If a user is accidentally assigned as a user manager or the opposite, the invitation can be revoked, and the user can be re-invited with the correct role.

### Before changes:
Screenshot 2023-07-07 at 12 02 02

### After changes:
Screenshot 2023-07-07 at 12 00 12

[CRIMRE-379]: https://dsdmoj.atlassian.net/browse/CRIMRE-379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


arthurashman (approved):
🔥 🎉

spec/system/manage_users/deactivate_a_user_spec.rb
```diff
@@ -98,9 +98,10 @@ def do_deactivate_journey
   end
 
   describe 'with only 2 active admins' do
+    let(:user_can_manage_others) { true }
```
arthurashman, line 101: I think this is already defaulted to true in the shared context no?

timpeat: One is the user, other is the current_user. I think name could be improved to avoid confusion.

arthurashman: Yeah, I do find the naming with these users confusing. 

Gotcha, I was looking at with_an_existing_user in main and forgot that you'd added the can_manage_others_attribute 👍

PR #343, opened 2023-07-07
Explicit configure of Sentry as error subscriber

timpeat:
## Description of change
Configure Sentry as a rails error subscriber

## Notes for reviewer

As discovered by @zheileman, and contrary to what is stated in https://edgeguides.rubyonrails.org/error_reporting.html, Sentry requires explicit configuration in order to register as a subscriber. This can be seen in the discussion on GitHub at https://github.com/rails/rails/pull/43625#issuecomment-1072514175.


zheileman (approved):
Nice 😃

PR #346, opened 2023-07-11
Snyk to run on chron job rather every branch / PR

willmcb:
## Description of change
Change so that Snyk runs on chron rather than every branch - should make the pipline a little faster.


zheileman (approved):
👍 
I think it might be good to have somewhere, maybe on confluence, instructions on how to access these reports?

PR #347, opened 2023-07-11
Changes the reply to address for access granted emails

willmcb:
## Description of change
Changes the reply to email address for access granted emails.

These have been added to notify for staging / prod - they are referenced like templates via UUID in:

`config/govuk_notify_templates.yml`

Note this only changes reply to for onboarding as the return email will use the default address - see ticket for more detail.

## Link to relevant ticket

[CRIMRE-376](https://dsdmoj.atlassian.net/browse/CRIMRE-376)

## Notes for reviewer
- please double check that the UUIDs are correct to your understanding. i.e. they match the UUID for the return to for each environment.

[CRIMRE-376]: https://dsdmoj.atlassian.net/browse/CRIMRE-376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/mailers/notify_mailer.rb
+++ b/app/mailers/notify_mailer.rb
@@ -31,6 +31,7 @@ class NotifyMailer < GovukNotifyRails::Mailer
 
   def access_granted_email(email)
     set_template(:access_granted_email)
+    set_email_reply_to(:onboarding_reply_to_address)
 
     mail(to: email)
   end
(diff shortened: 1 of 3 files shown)


timpeat (approved):
👍

PR #351, opened 2023-07-13
CRIMRE-402 ensure dev_auth cannot be used outside local env

timpeat:
## Description of change
Prevent the DevAuth strategy from being configured outside the local env

## Link to relevant ticket
[CRIMRE-402](https://dsdmoj.atlassian.net/browse/CRIMRE-402)

## Notes for reviewer
Adds an additional check to prevent the application being initialised with the DevAuth feature enabled in a non Local environment.

[CRIMRE-402]: https://dsdmoj.atlassian.net/browse/CRIMRE-402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -36,6 +36,10 @@ Devise.setup do |config|
   #
   # Uses the DevAuth strategy if local and ENV["DEV_AUTH_ENABLED"] is true
 
+  if !HostEnv.local? && FeatureFlags.dev_auth.enabled?
+    raise "The DevAuth strategy must not be used in this environment"
+  end
+
   strategy_class = FeatureFlags.dev_auth.enabled? ? OmniAuth::Strategies::DevAuth : OmniAuth::Strategies::OpenIDConnect
 
   config.omniauth(
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -1,12 +1,8 @@
 feature_flags:
-  developer_tools:
-    local: true
-    staging: true
-    production: false
   dev_auth:
     local: <%= ENV.fetch("DEV_AUTH_ENABLED", false) %>
     staging: false
-    production: false
+    production: false # must never be enabled in the live service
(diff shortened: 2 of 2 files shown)


config/settings.yml
```diff
@@ -1,12 +1,8 @@
 feature_flags:
-  developer_tools:
-    local: true
-    staging: true
-    production: false
   dev_auth:
     local: <%= ENV.fetch("DEV_AUTH_ENABLED", false) %>
     staging: false
-    production: false
+    production: false # must never be enabled in the live service
```
willmcb, line 5: 👍

config/initializers/devise.rb
```diff
@@ -36,6 +36,10 @@
   #
   # Uses the DevAuth strategy if local and ENV["DEV_AUTH_ENABLED"] is true
 
+  if !HostEnv.local? && FeatureFlags.dev_auth.enabled?
+    raise "The DevAuth strategy must not be used in this environment"
```
willmcb, line 40: Wondering if it is worth having a specific exeption class for this as well? Might be easier to spot - small thing though.

willmcb (approved):
Thanks for this 👍

config/initializers/devise.rb
mseedat-moj: Depends how much faith we have in `HostEnv.local?` being set correctly? I don't think there are many ways of checking whether an environment is considered local or not. Perhaps a check somewhere on the incoming HTTP Host name - if it looks remotely unproduction-link then flash up a warning/exception somewhere asap? i don't think there's a way of knowing what the full server URL is on system initialize?

timpeat: @willmcb the error will appear in the build history and we won't be doing anything else with it so I think just setting the message is fine.

PR #362, opened 2023-07-17
CRIMRE-411 ignore open reviews in caseworker filter

timpeat:
## Description of change
Only closed reviews have a reviewer. Open Reviews need to be excluded when building the list of reviewers for the search dropdown.

## Notes for reviewer

This was causing a user with an id of nil being added to the search filter dropdown. When deployed to staging, which has some in progress reviews,  a user called '[deleted]' with an id of nil appeared in the caseworker filter dropdown.

### Before changes:
Screenshot 2023-07-17 at 16 13 51

### After changes:
Screenshot 2023-07-17 at 16 14 01

### DIFF
--- a/app/models/application_search_filter.rb
+++ b/app/models/application_search_filter.rb
@@ -56,11 +56,14 @@ class ApplicationSearchFilter < ApplicationStruct
   private
 
   def assigned_to_user_options
-    user_ids = (
-      CurrentAssignment.distinct.pluck(:user_id) + Review.distinct.pluck(:reviewer_id)
-    ).uniq
+    # builds an array of [User#name, User#id] sorted by first name, last name
+    ids_of_users_with_active_assignments_and_or_reviews.map { |id| [User.name_for(id), id] }.sort
+  end
 
-    user_ids.map { |id| [User.name_for(id), id] }.sort
+  def ids_of_users_with_active_assignments_and_or_reviews
+    (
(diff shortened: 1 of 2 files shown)


app/models/application_search_filter.rb
```diff
@@ -56,11 +56,14 @@ def datastore_params
   private
 
   def assigned_to_user_options
-    user_ids = (
-      CurrentAssignment.distinct.pluck(:user_id) + Review.distinct.pluck(:reviewer_id)
-    ).uniq
+    # builds an array of [User#name, User#id] sorted by first name, last name
+    ids_of_users_with_active_assignemnts_and_or_reviews.map { |id| [User.name_for(id), id] }.sort
```
arthurashman, line 60: sp. on assignments

app/models/application_search_filter.rb
```diff
@@ -56,11 +56,14 @@ def datastore_params
...
+  end
 
-    user_ids.map { |id| [User.name_for(id), id] }.sort
+  def ids_of_users_with_active_assignemnts_and_or_reviews
```
arthurashman, line 63: sp. on assignments

app/models/application_search_filter.rb
timpeat: gah, good call. More haste, less speed!

arthurashman (approved):
👍

PR #363, opened 2023-07-18
CRIMRE-415 extend the workload report by one day

timpeat:
## Description of change
Add a dedicated row for "3 days" in the workload report.

## Link to relevant ticket
[CRIMRE-415](https://dsdmoj.atlassian.net/browse/CRIMRE-415)

### After changes:
image

[CRIMRE-415]: https://dsdmoj.atlassian.net/browse/CRIMRE-415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/models/reporting/workload_report_spec.rb
+++ b/spec/models/reporting/workload_report_spec.rb
@@ -1,7 +1,7 @@
 require 'rails_helper'
 
 RSpec.describe Reporting::WorkloadReport do
-  subject(:report) { described_class.new day_zero: }
+  subject(:report) { described_class.new day_zero: day_zero, number_of_rows: 4 }
 
   let(:day_zero) { Date.parse('2023-01-04') }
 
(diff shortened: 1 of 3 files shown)


spec/models/reporting/workload_report_spec.rb
```diff
@@ -68,11 +68,7 @@
     end
   end
 
-  describe 'number of rows can be configured using :number_of_days' do
-    subject(:report) { described_class.new(number_of_rows: 7) }
-
-    it 'returns the counts for the four working days up to and including day_zero' do
-      expect(report.table.rows.size).to be 7
-    end
+  it 'number of rows defaults to 5' do
```
willmcb, line 71: Just thinking is this not a different test than a replacement? As in, it can be configured is one thing, and the fact it is defaulted to 5 is another.

Could we have both tests? or do you think that is overkill?

willmcb (approved):
LGTM - tiny comment to consider though 👍

spec/models/reporting/workload_report_spec.rb
timpeat: I've updated the other specs to use a configured rather than default value, so that case is handled there. This is just for the default.

spec/models/reporting/workload_report_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'rails_helper'
 
 RSpec.describe Reporting::WorkloadReport do
-  subject(:report) { described_class.new day_zero: }
+  subject(:report) { described_class.new day_zero: day_zero, number_of_rows: 4 }
```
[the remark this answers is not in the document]
timpeat: I've locked it down as 4 in the unit tests otherwise we'd have to keep re-writing them when the number of rows change.

PR #364, opened 2023-07-18
[CHORE] Clean translation files

mseedat-moj:
## Description of change
Removes whitespaces which causes problems with diffs

## Link to relevant ticket
N/A


arthurashman (approved):
Thanks @mseedat-moj !

PR #368, opened 2023-07-20
Fix docker compose

hiboabd:
## Description of change
This PR fixes running docker-compose in review. The docker-compose env name has been changed to staging to enable dev_auth to be used (as dev_auth cannot be used in production) and the running of the app in ssl has been disabled

## Link to relevant ticket
[CRIMRE-370](https://dsdmoj.atlassian.net/browse/CRIMRE-370)

## Notes for reviewer
Fyi I found testing in chrome to be a bit of a pain as it kept forcing ssl. I found Firefox to be easier 

[CRIMRE-370]: https://dsdmoj.atlassian.net/browse/CRIMRE-370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,9 @@ The last user in the list is "Not.Authorised@example.com." Select this user to s
 
 This authentication strategy can be disabled locally by setting ```DEV_AUTH_ENABLED=false```
 
+**NOTE:** The dev_auth authentication strategy has been enabled for local development and docker-compose by setting ```DEV_AUTH_ENABLED=true``` and ```IS_LOCAL_DOCKER_ENV=true```.
+This must never be enabled in the live (staging/production) service in it's current form as it will expose sensitive information.
+
 ## Local development with the temporary API gem
 
 The app utilises [puma-dev](https://github.com/puma/puma-dev) for local development.
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -48,7 +48,7 @@ Rails.application.configure do
   # config.action_cable.allowed_request_origins = [ "http://example.com", /http:\/\/example.*/ ]
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = true
+  config.force_ssl = ENV.key?('DISABLE_HTTPS') ? false : true
 
   # Include generic and useful information about system operation, but avoid logging too much
   # information to avoid inadvertent exposure of personally identifiable information (PII).
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -34,10 +34,10 @@ Devise.setup do |config|
   # Add a new OmniAuth provider. Check the wiki for more information on setting
   # up on your models and hooks.
   #
-  # Uses the DevAuth strategy if local and ENV["DEV_AUTH_ENABLED"] is true
+  # Uses the DevAuth strategy if local/docker env and ENV["DEV_AUTH_ENABLED"] is true
 
-  if HostEnv.production? && FeatureFlags.dev_auth.enabled?
-    raise "The DevAuth strategy must not be used in this environment"
+  unless HostEnv.local? || ENV.key?('IS_LOCAL_DOCKER_ENV')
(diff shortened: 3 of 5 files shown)


config/initializers/devise.rb
```diff
@@ -36,7 +36,7 @@
   #
   # Uses the DevAuth strategy if local and ENV["DEV_AUTH_ENABLED"] is true
 
-  if !HostEnv.local? && FeatureFlags.dev_auth.enabled?
+  if HostEnv.production? && FeatureFlags.dev_auth.enabled?
     raise "The DevAuth strategy must not be used in this environment"
```
hiboabd, line 40: Technically this error message is correct but is it worth being specific? - "The DevAuth strategy must not be used in production"

config/environments/production.rb
```diff
@@ -48,7 +48,7 @@
   # config.action_cable.allowed_request_origins = [ "http://example.com", /http:\/\/example.*/ ]
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = true
+  config.force_ssl = ENV.key?('DISABLE_HTTPS') ? false : true
```
mseedat-moj, line 51: this line scares me

zheileman: Should be fine. In cloud deployed apps, anything coming through the ingress will redirect to https, regardless of this rails config flag. But is important to be able to bypass it for docker-compose.

https://kubernetes.github.io/ingress-nginx/user-guide/tls/#server-side-https-enforcement-through-redirect

> By default the controller redirects HTTP clients to the HTTPS port 443 using a 308 Permanent Redirect response if TLS is enabled for that Ingress.

config/settings.yml
```diff
@@ -1,7 +1,7 @@
 feature_flags:
   dev_auth:
     local: <%= ENV.fetch("DEV_AUTH_ENABLED", false) %>
-    staging: false
+    staging: <%= ENV.fetch("DEV_AUTH_ENABLED", false) %>
```
zheileman, line 4: Is this staging feature flag still needed or with the change in devise initialiser it will never be possible to enable dev_auth on staging (I think as expected to avoid revealing users?)

docker-compose.yml
```diff
@@ -14,7 +14,7 @@ services:
   web:
     build: .
     environment:
-      ENV_NAME: production
```
zheileman, line 17: Could this now be `production`? As we've introduced the `IS_LOCAL_DOCKER_ENV` variable anyways I think it would work?

README.md
```diff
@@ -51,6 +51,9 @@ The last user in the list is "Not.Authorised@example.com." Select this user to s
 
 This authentication strategy can be disabled locally by setting ```DEV_AUTH_ENABLED=false```
 
+**NOTE:** The dev_auth authentication strategy has been enabled for local development and docker-compose by setting ```DEV_AUTH_ENABLED=true``` and ```IS_LOCAL_DOCKER_ENV=true```.
+This must never be enabled in the live (staging/production) service in it's current form as it will expose sensitive information.
```
hiboabd, line 55: I tried to be a bit vague as to what would be exposed as it doesn't seem like a good idea to shout about it in the readme 😅 is it worth keeping this?

config/settings.yml
hiboabd: It's still needed as it sets the the dev_auth feature flag to true. The change in the devise initialiser to put the host env local check back in means it can't be enabled on staging

docker-compose.yml
hiboabd: It could be but I would need to change the settings.yml config to ` production: <%= ENV.fetch("DEV_AUTH_ENABLED", false) %>` which i was bit reluctant to do

config/settings.yml
zheileman: Ah sorry yes, understood. Disregard the other comment as I see now these 2 changes go in tandem 👍

PR #369, opened 2023-07-20
Add 422 error message

timpeat:
## Description of change
Previously invalid authenticity token requests etc were being shown a page not found message. This change shows a request error message suggesting they go back, refresh, and try again.

Also overrides Devise passthru to redirect to sign in. Otherwise a user gets stuck on an unformatted page if they refresh the after an authenticity token failure on the user auth path.

### Before changes:
Screenshot 2023-07-20 at 13 18 19

### After changes:

Screenshot 2023-07-20 at 16 13 04

From the root path, tamper with the authenticity token hidden field on the Start button, and click the button.
To confirm the passthru fix, reload the page and confirm you are redirected to sign_in.

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -13,5 +13,13 @@ module Users
     def failure
       throw(:warden, recall: 'Errors#forbidden', message: :forbidden)
     end
+
+    # Override the #passthru action. It is used when a GET request is made
+    # to the user auth url. Ideally the GET route would not be added by Devise.
+    # The fix for this is in Devise but awaiting release:
+    # https://github.com/heartcombo/devise/pull/5508
+    def passthru
+      redirect_to unauthenticated_root_path
+    end
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -13,5 +13,13 @@ def azure_ad
     def failure
       throw(:warden, recall: 'Errors#forbidden', message: :forbidden)
     end
+
+    # Override the #passthru action. It is used when a GET request is made
+    # to the user auth url. Ideally the GET route would not be added by Devise.
+    # The fix for this is in Devise but awaiting release:
+    # https://github.com/heartcombo/devise/pull/5508
```
willmcb, line 20: Looks like they release fairly often - shall we add a slack reminder in on the dev channel to check this in say a month so that we don't forget it?

willmcb (approved):
LGTM 👍

app/controllers/users/omniauth_callbacks_controller.rb
timpeat: I have an alert set :). Watched kettle never boils and all that :)

PR #376, opened 2023-07-28
CRIMAP-499 filter by age in business days.

timpeat:
## Description of change
Users can search applications by age in business days.
This feature is currently only enabled in local environments.

## Link to relevant ticket
[CRIMAP-499](https://dsdmoj.atlassian.net/browse/CRIMAP-499)

## Notes for reviewer
The code surrounding the filter is more complex than I'd like. Most of this complexity arises because the data source for the search changes depending on which filters are active. "Caseworker" and "Business days since received" both use local data, while the remaining filters are for the datastore.

### Without the feature flag enabled:
Screenshot 2023-07-28 at 14 04 04

### With the feature flag enabled:
Screenshot 2023-07-28 at 14 03 30

[CRIMAP-499]: https://dsdmoj.atlassian.net/browse/CRIMAP-499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/application_search.rb
+++ b/app/models/application_search.rb
@@ -6,6 +6,10 @@ class ApplicationSearch
   end
 
   def results
+    # If the result set from the local Review database query is empty, we know we do not need to perform a
+    # datastore search.
+    return [] if @filter.datastore_results_will_be_empty?
+
     @results ||= datastore_search_response.map do |result|
       application = ApplicationSearchResult.new(result)
 
--- a/app/models/application_search_filter.rb
+++ b/app/models/application_search_filter.rb
@@ -1,19 +1,20 @@
 class ApplicationSearchFilter < ApplicationStruct
-  attribute? :assigned_status, Types::AssignedStatus | Types::Uuid
+  REVIEW_FILTERS = %i[assigned_status age_in_business_days].freeze
+  DATASTORE_FILTERS = %i[
+    applicant_date_of_birth application_id_in application_status search_text submitted_after submitted_before
+  ].freeze
+
+  attribute? :age_in_business_days, Types::Params::Nil | Types::Params::Integer
+  attribute? :applicant_date_of_birth, Types::Params::Nil | Types::Params::Date
   attribute? :application_status, Types::ReviewStatusGroup
+  attribute? :assigned_status, Types::AssignedStatus | Types::Uuid
   attribute? :search_text, Types::Params::Nil | Types::Params::String
   attribute? :submitted_after, Types::Params::Nil | Types::Params::Date
   attribute? :submitted_before, Types::Params::Nil | Types::Params::Date
-  attribute? :applicant_date_of_birth, Types::Params::Nil | Types::Params::Date
 
-  #
   # Options for the assigned status filter
-  #
   # Includes "Unassigned", "All assigned" prepended to a list of all user names.
   # Values can be "unassigned", "assigned" or a user id.
-  # If a user is specified, only applications assinged to or reviewed by that
-  # user are returned.
-  #
   def assigned_status_options
     status_options = Types::ASSIGNED_STATUSES.map do |status|
       [I18n.t(status, scope: 'values.assigned_status'), status]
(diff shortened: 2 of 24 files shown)


app/models/application_search.rb
```diff
@@ -6,6 +6,9 @@ def initialize(filter:, pagination:, sorting: Sorting.new)
   end
 
   def results
+    # If the active search filters guarantee empty results, we skip the datastore query.
+    return [] if @filter.datastore_results_will_be_empty?
```
willmcb, line 10: 👍

app/models/application_search_filter.rb
```diff
@@ -22,96 +21,116 @@ def assigned_status_options
...
-    }
+    active_datastore_filters.each_with_object({}) do |filter, params|
+      params.merge!(send("#{filter}_datastore_param"))
+    end
+  end
+
+  # If there are no applications that satisfy the active review filters,
+  # there is no need to perform a datastore query as the results will be empty.
+  def datastore_results_will_be_empty?
+    !active_review_filters.empty? && active_review_filters_application_ids.empty?
```
willmcb, line 51: Can you explain how this works to me?

app/models/application_search_filter.rb
```diff
@@ -22,96 +21,116 @@ def assigned_status_options
...
-  def application_id_in
-    case assigned_status
-    when nil, 'unassigned'
-      []
-    when 'assigned'
-      all_assigned_application_ids
-    else
-      application_ids_for_user(assigned_status)
-    end
+  def active_review_filters
```
willmcb, line 60: Thinking this might be more clear as `unique_active_review_filters` - if they can exist in both active filters and REVIEW_FILTERS they I understand the need for `&` if they cant then better to just concatenate the arrays.

app/models/application_search_filter.rb
```diff
@@ -22,96 +21,116 @@ def assigned_status_options
...
   end
 
-  def application_ids_for_user(user_id)
-    CurrentAssignment.assigned_to_ids(user_id:) +
-      Review.reviewed_by_ids(user_id:)
+  # Returns an array of the active datastore filters that will be used in the datastore search.
+  # If there are active review filters, they are included in this list as ":application_id_in".
+  def active_datastore_filters
+    filters = (active_filters & DATASTORE_FILTERS)
+    filters << :application_id_in unless active_review_filters.empty?
```
willmcb, line 68: Why not put the logic for checking `active_review_filters` in that method. Seems like the comment is describing what the output of this method should be.

config/environments/test.rb
```diff
@@ -24,6 +24,7 @@
 
   # Show full error reports and disable caching.
   config.consider_all_requests_local       = false
+
```
willmcb, line 27: Is there a reason to add this space?

config/locales/en.yml
```diff
@@ -88,58 +88,58 @@ en:
     manage_users: Manage users
 
   labels: &LABELS
-    select_return_reason: Select the reason for returning this application
-    return_reason_details_legend: Give further details
-    return_reason_details: We'll share this information with the provider
+    age_in_business_days: "Business days since received"
```
willmcb, line 91: Don't think there is a need for the double quotes here.

config/locales/en.yml
```diff
@@ -88,58 +88,58 @@ en:
...
+    offence: Offence
+    offence_date: Offence date
+    office_account_number: Office account number
+    other_names: Other names
+    overall_offence_class: Overall offence class
+    partner: Partner
+    previous_maat_id: Previous MAAT ID
     provider_email: Email address
+    provider_phone_number: Phone
+    return_reason_details: We'll share this information with the provider
```
willmcb, line 131: I think perhaps avoid the conjuction "We'll" --> "We will" is just a bit clearer

willmcb (approved):
Looks good I don't think there is anything blocking here - would be good to have a bit of a chat with you about id if you have some time though.

PR #377, opened 2023-08-01
[CRIMAP-508/512/513] Introduce basic user role change journey

mseedat-moj:
## Description of change
Introduces basic user roles implementation, starting with `caseworker` and `supervisor`. Feature is hidden by a Feature Flag. Shows change in user history. Shows current role in Manager Users-> active users table.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-508 (completes UI part)
https://dsdmoj.atlassian.net/browse/CRIMAP-512
https://dsdmoj.atlassian.net/browse/CRIMAP-513

### Raised:
https://dsdmoj.atlassian.net/browse/CRIMAP-531 - move user actions
https://dsdmoj.atlassian.net/browse/CRIMAP-532 - notify via email
https://dsdmoj.atlassian.net/browse/CRIMAP-533 - UX and Content review

## Notes for reviewer
1. ~~There is an existing `Types::UserRole` which I initially removed but then re-introduced to act as a source of truth and consistency when creating the `ChangeRole` attributes `to:` and `from` which would otherwise be `Types::String`~~ resolved in #379 
2. ~~There is a new module called `UserRole` as well - is that the right naming for this concern - others have `_able` style e.g. `Capable`, `Rollable` or even `Rollify` etc!~~ resolved in #379 
3. ~~There is a rule added to ensure `dormant` users cannot have their role changed - is that correct?~~  resolved in #379 
4. There is a new ticket CRIMAP-531 to move the user `actions` into the current user history page to give more space/options
5. The command `ChangeRole` does not decide the new role at the moment even though there are only 2, instead the controller provides the 'new role' based on the assumption that if there are more roles you can't just toggle between 2 of them. Is that a wise assumption or too presumptive?
6. ~~The User.role value is an enum, therefore roles cannot be stacked. Is that ok?~~ resolved in #379 
7. The command `ChangeRole` has guard clauses - is this the right place to put them?

### Before changes:
N/A

### After changes:

Screenshot 2023-08-02 at 16 44 03

#### Interstitial page when role change initiated
Screenshot 2023-08-01 at 10 18 21

#### Role toggled from Supervisor to Caseworker successfully, also shows 'change role' action
Screenshot 2023-08-01 at 10 18 33

#### User history shows changed role
Screenshot 2023-08-01 at 10 18 57

#### 

1. Ensure `settings.yml` has the `basic_user_roles` value for the test environment set to `true`
2. Log in as an `can manage users` user, navigate to Active Users
3. Change existing user roles as desired


app/models/concerns/revivable.rb
```diff
@@ -12,7 +12,7 @@ def dormant?
   end
 
   def revivable?
-    activated? && last_auth_at < Rails.configuration.x.auth.dormant_account_threshold.ago
+    activated? && last_auth_at.present? && last_auth_at < Rails.configuration.x.auth.dormant_account_threshold.ago
```
mseedat-moj, line 15: Added the last_auth_at.present? check to prevent nil exceptions, especially during testing.

timpeat (approved):
Looks Great! 🚀

willmcb (approved):
LGTM 👍

PR #381, opened 2023-08-04
[CRIMAP-509/510] Add supervisor performance tracking page

hiboabd:
## Description of change
Adds performance tracking page `/performance_tracking` which is hidden behind the `BasicUserRoles` feature flag and access is restricted to supervisors only.
Upon hitting the url, caseworkers are shown a 404 page and user managers are redirected to the manage users dashboard. The tab is also hidden from both user types. 

## Link to relevant ticket
[CRIMAP-509](https://dsdmoj.atlassian.net/browse/CRIMAP-509)
[CRIMAP-510](https://dsdmoj.atlassian.net/browse/CRIMAP-510)

## Notes for reviewer
Not a fan of the Performance tracking name 😅 I'm assuming this is just a placeholder but other names are welcome 😄
Also now that we are introducing new roles and routes, it is worth potentially rethinking the current `authentication_spec.rb` format as I can see it becoming a big file down the line. Perhaps the testing of which areas of the service each role has access to could be split into different files?  

### After changes:

**Supervisor view** 
Screenshot 2023-08-04 at 11 51 44

**Caseworker view**
Screenshot 2023-08-04 at 11 52 26

Screenshot 2023-08-04 at 11 52 37

**User manager view**

Screenshot 2023-08-04 at 11 52 12

## How to manually test the feature

[CRIMAP-509]: https://dsdmoj.atlassian.net/browse/CRIMAP-509?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CRIMAP-510]: https://dsdmoj.atlassian.net/browse/CRIMAP-510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #383, opened 2023-08-07
CRIMAP-535-User manager has access to performance tracking page

MohammedHassanLatif93:
## Description of change
User manager has access to performance tracking page on staging only

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-535

## Notes for reviewer
This is how we would expect this to look in staging these screenshots were obtained when setting   allow_user_managers_service_access to true.

### Before changes:
image

### After changes:
image

### DIFF
--- a/app/controllers/performance_tracking_controller.rb
+++ b/app/controllers/performance_tracking_controller.rb
@@ -7,7 +7,6 @@ class PerformanceTrackingController < ServiceController
 
   # Users that can manage others are only allowed to access this page on staging.
   # This is configured by the allow_user_managers_service_access feature flag.
-
   def require_supervisor!
     return if current_user.supervisor? && FeatureFlags.basic_user_roles.enabled?
     return if FeatureFlags.allow_user_managers_service_access.enabled?
--- a/app/controllers/performance_tracking_controller.rb
+++ b/app/controllers/performance_tracking_controller.rb
@@ -5,8 +5,12 @@ class PerformanceTrackingController < ServiceController
 
   private
 
+  # Users that can manage others are only allowed to access this page on staging.
+  # This is configured by the allow_user_managers_service_access feature flag.
+
   def require_supervisor!
     return if current_user.supervisor? && FeatureFlags.basic_user_roles.enabled?
+    return if FeatureFlags.allow_user_managers_service_access.enabled?
 
     render status: :not_found, template: 'errors/not_found', layout: 'errors'
     false
--- a/app/controllers/performance_tracking_controller.rb
+++ b/app/controllers/performance_tracking_controller.rb
@@ -8,7 +8,7 @@ class PerformanceTrackingController < ServiceController
   # Users that can manage others can access this on staging for testing purposes.
   def require_supervisor!
     return if current_user.supervisor? && FeatureFlags.basic_user_roles.enabled?
-    return if FeatureFlags.allow_user_managers_service_access.enabled?
+    return if allow_user_managers_service_access?
 
     render status: :not_found, template: 'errors/not_found', layout: 'errors'
     false
--- a/app/controllers/performance_tracking_controller.rb
+++ b/app/controllers/performance_tracking_controller.rb
@@ -1,0 +1,14 @@
+class PerformanceTrackingController < ServiceController
+  before_action :require_supervisor!
+
+  def index; end
+
+  private
+
+  def require_supervisor!
+    return if current_user.supervisor? && FeatureFlags.basic_user_roles.enabled?
+
+    render status: :not_found, template: 'errors/not_found', layout: 'errors'
+    false
+  end
+end
--- a/app/controllers/performance_tracking_controller.rb
+++ b/app/controllers/performance_tracking_controller.rb
@@ -5,8 +5,7 @@ class PerformanceTrackingController < ServiceController
 
   private
 
-  # Users that can manage others are only allowed to access this page on staging.
-  # This is configured by the allow_user_managers_service_access feature flag.
+  # Users that can manage others can access this on staging for testing purposes.
   def require_supervisor!
     return if current_user.supervisor? && FeatureFlags.basic_user_roles.enabled?
     return if FeatureFlags.allow_user_managers_service_access.enabled?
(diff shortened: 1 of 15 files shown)


app/controllers/performance_tracking_controller.rb
```diff
@@ -5,8 +5,11 @@ def index; end
 
   private
 
+  # Users that can manage others are only allowed to access this page on staging.
+  # This is configured by the allow_user_managers_service_access feature flag.
   def require_supervisor!
```
mseedat-moj, line 10: I would remove this comment as the `FeatureFlags.allow...` line of code says the same thing imo.

app/controllers/performance_tracking_controller.rb
```diff
@@ -5,8 +5,11 @@ def index; end
...
     return if current_user.supervisor? && FeatureFlags.basic_user_roles.enabled?
+    return if FeatureFlags.allow_user_managers_service_access.enabled?
 
     render status: :not_found, template: 'errors/not_found', layout: 'errors'
```
mseedat-moj, line 14: Line 14 is never reached at the moment - is that intentional?

app/views/layouts/_header.html.erb
```diff
@@ -2,7 +2,8 @@
   <%= header.with_navigation_item(text: current_user.name) %>
   <% if current_user.can_manage_others %>
     <%= header.with_navigation_item(text: t('.manage_users'), href: admin_manage_users_root_path) %>
-  <% elsif current_user.supervisor? && FeatureFlags.basic_user_roles.enabled? %>
+  <%end%>
```
mseedat-moj, line 5: `<% end %>` (spaces!) - does the `elsif` need to go i.e. can both links be shown simultaneously?

spec/system/performance_tracking/viewing_spec.rb
```diff
@@ -42,5 +42,20 @@
         expect(heading_text).to eq('Performance tracking')
       end
     end
+
+    context 'when user managers are logged in on staging' do
+      before do
+        # Override allow_user_managers_service_access as per staging
```
mseedat-moj, line 48: No need to comment here imo, Line 49 is fairly self-explanatory imo. Use comments to justify code decisions/explain _why_ something works in a particular way, let the code explain _how_ that was done

mseedat-moj (changes requested):

app/views/layouts/_header.html.erb
```diff
@@ -2,7 +2,8 @@
...
+  <% if (current_user.supervisor? && FeatureFlags.basic_user_roles.enabled?) || FeatureFlags.allow_user_managers_service_access.enabled? %>
```
willmcb, line 6: This is a lot of logic to have a in a view - I think we could wrap it in a method to encapsulate it and make the view more readable 👍

spec/system/header_nav_spec.rb
```diff
@@ -80,6 +80,12 @@
                                                    page.first('.govuk-heading-xl').text
                                                  }.from('Your list').to('Manage users')
       end
+
+      it 'shows the "Performance tracking" link and can follow it' do
```
willmcb, line 84: Small point but I think it would make more sense to say 'shows the "Performance tracking" link and it can be followed'

spec/system/performance_tracking/viewing_spec.rb
```diff
@@ -42,5 +42,20 @@
         expect(heading_text).to eq('Performance tracking')
       end
     end
+
+    context 'when user managers are logged in on staging' do
```
willmcb, line 46: Nice 👍

app/controllers/performance_tracking_controller.rb
```diff
@@ -5,8 +5,11 @@ def index; end
 
   private
 
+  # Users that can manage others are only allowed to access this page on staging.
+  # This is configured by the allow_user_managers_service_access feature flag.
   def require_supervisor!
     return if current_user.supervisor? && FeatureFlags.basic_user_roles.enabled?
+    return if FeatureFlags.allow_user_managers_service_access.enabled?
```
willmcb, line 12: This logic is repeated in the view. We could add a helper method in the controller so that it can be used here and in the controller.

https://apidock.com/rails/ActionController/Helpers/ClassMethods/helper_method

Would be good to write a test for this as well that exercises all of the possible options that are expressed in the logic here.

User stubbing to make change the feature flags + users that have different roles to check it.

willmcb (changes requested):
Nice work, almost there - have a look at the comments, happy to chat them through with you tomorrow if you'd like :)

app/views/layouts/_header.html.erb
hiboabd: Ah I think hassan's ide doesn't format automatically 😅 And yes for user managers they would see both links on staging. Supervisors can't manage others so they would only see the performance tracking tab

spec/system/performance_tracking/viewing_spec.rb
willmcb: Agree with this.

app/controllers/performance_tracking_controller.rb
hiboabd: What do you mean never reached? It is reached for caseworkers who shouldn't be able to see the page

app/controllers/performance_tracking_controller.rb
hiboabd: We'll could adjust the comment to explain why user managers can have access in staging (i.e. for testing purposes). I agree the second line isn't necessary

app/views/layouts/_header.html.erb
hiboabd: We did discuss this and I agree, however the `basic_user_roles` feature flag is not going to remain and will be removed when the user roles work goes live so it didn't feel necessary in the long run to encapsulate it into it's own method...  happy to add a method however

app/views/layouts/_header.html.erb
hiboabd: We just ran `bundle exec erblint app/views/ -a` and it's not catching it 🤔

app/controllers/performance_tracking_controller.rb
hiboabd: We do have tests that cover these different scenarios from [our previous pr](https://github.com/ministryofjustice/laa-review-criminal-legal-aid/pull/381). We only added the new scenarios (when a user manager can access the page on staging) in this PR. But happy to chat through

app/controllers/performance_tracking_controller.rb
mseedat-moj: Ah sorry misread the statement!

mseedat-moj (approved):
👍

PR #386, opened 2023-08-09
Ensure only user managers can access parts of the service on staging

hiboabd:
## Description of change
The service and performance tracking page are accessible to user managers by enabling the `allow_user_managers_service_access` feature flag on staging. However, whether the feature flag was enabled was the only condition being checked in the controllers meaning every user has access to the feature on staging as the feature flag is enabled for all on staging. This was not obvious previously as the service is also accessible to caseworkers and supervisors.

This condition has been changed to check whether a user can manage others as well as checking that the feature flag is enabled and tests have been modified to check both caseworker and user manager access of the performance tracking page on staging

## Link to relevant ticket
[CRIMAP-535](https://dsdmoj.atlassian.net/browse/CRIMAP-535)

## Notes for reviewer
Merging this PR initially to prevent merging being blocked, will follow up with a pr to refactor specs 

### Before changes:
Currently on staging the performance tracking tab is accessible to caseworkers when it shouldn't be 
Screenshot 2023-08-09 at 10 05 40

[CRIMAP-535]: https://dsdmoj.atlassian.net/browse/CRIMAP-535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -8,7 +8,7 @@ class ApplicationController < ActionController::Base
   end
 
   def allow_user_managers_service_access?
-    FeatureFlags.allow_user_managers_service_access.enabled?
+    current_user.can_manage_others? && FeatureFlags.allow_user_managers_service_access.enabled?
   end
 
   # Redirect user managers to manage users' admin on sign in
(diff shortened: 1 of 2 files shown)


mseedat-moj (approved):
From a code perspective seems fine for now but things are starting to get complicated quickly

PR #387, opened 2023-08-09
Use separate in memory event store for each spec

timpeat:
## Description of change

Use separate in-memory event store for each spec

## Notes for reviewer

I'd originally made this change as part of the caseworker report feature but have extracted it here as it is a distinct change-set.

The introduction of the in-memory event store leads to a decrease in the frequency of database queries, resulting in accelerated test execution (approximately 10% faster for model specs on my M1).

Using distinct event stores for each spec, as opposed to the one globally configured by the event_store initialiser means that handlers are not subscribed by default. This means you can include only the required handler subscriptions based on the specific scenario under test. For instance, while testing reporting functionality, event creation becomes possible without the necessity of managing email notifiers or other subscribers.


spec/models/reporting/received_on_report_spec.rb
```diff
@@ -1,6 +1,12 @@
 require 'rails_helper'
 
 RSpec.describe Reporting::ReceivedOnReport do
+  before do
+    # Models specs do not have handers subscribed to the event store by default.
+    # Here we subscribe the RecievedOnReports handlers to the event store
```
willmcb, line 6: Sorry to be a pedant but there's a typo here: 

`Rec**ei**vedOnReports`

👍

spec/rails_helper.rb
```diff
@@ -53,6 +52,25 @@
...
+      repository: RubyEventStore::InMemoryRepository.new
+    )
+  end
+
+  # Subscribe all apart from notifier handler for system specs
+  config.before(:each, type: :system) do |_example|
+    event_store = Rails.configuration.event_store
+
+    CurrentAssignments::Configuration.new.call(event_store)
+    # ReceivedOnReports::Configuration.new.call(event_store)
```
willmcb, line 71: Is there a reason you have left this commented out?

willmcb (changes requested):
Nice one - some small things to look at :)

spec/rails_helper.rb
timpeat: sorry. fixed

willmcb (approved):
👍 Nice one, such an improvement :)

PR #389, opened 2023-08-10
[CRIMAP-553] Turn off ssl in localhost

hiboabd:
## Description of change
Disables force_ssl in local development. This is currently enabled in `application.rb` for user authentication. 

## Link to relevant ticket
[CRIMAP-553](https://dsdmoj.atlassian.net/browse/CRIMAP-553)

### Before changes:
Screenshot 2023-08-10 at 14 20 08

### After changes:
Screenshot 2023-08-10 at 14 19 36

[CRIMAP-553]: https://dsdmoj.atlassian.net/browse/CRIMAP-553?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -73,4 +73,7 @@ Rails.application.configure do
 
   # Allow connections from inside a docker container to this host machine
   config.hosts += %w[host.docker.internal]
+
+  # Provides the option to turn off SSL in localhost
+  config.force_ssl = ENV.key?('DISABLE_HTTPS') ? false : true
 end
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -74,6 +74,5 @@ Rails.application.configure do
   # Allow connections from inside a docker container to this host machine
   config.hosts += %w[host.docker.internal]
 
-  # Provides the option to turn off SSL in localhost
-  config.force_ssl = ENV.key?('DISABLE_HTTPS') ? false : true
+  config.force_ssl = false
 end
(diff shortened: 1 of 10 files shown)


zheileman:
@hiboabd  But is there a reason why this is not the _default_ behaviour like with any other rails app, without having to set any ENV variables?
Does review need SSL on localhost?

hiboabd:
> @hiboabd But is there a reason why this is not the _default_ behaviour like with any other rails app, without having to set any ENV variables? Does review need SSL on localhost?

Yeah you're right it's not necessary i've removed it

config/environments/development.rb
```diff
@@ -73,4 +73,6 @@
 
   # Allow connections from inside a docker container to this host machine
   config.hosts += %w[host.docker.internal]
+
+  config.force_ssl = false
```
zheileman, line 77: I think this shouldn't be necessary. The default when running in development environment is false, no need to declare this.
So if it is only working by declaring it, there is something else going on. If you take a look at the datastore and apply `development.rb` we didn't have to add this 🤔

hiboabd: Yes I noticed but it's because it is set [in application.rb ](https://github.com/ministryofjustice/laa-review-criminal-legal-aid/blob/main/config/application.rb#L38) so it needs to be overridden

zheileman (approved):
Nice!

PR #390, opened 2023-08-10
Lock down permissions for access to performance tracking

arthurashman:
## Description of change
Ensure that should a user_manager have a Supervisor role in prod that they cannot see the Performance tracking nav link

Make user_manager permissions in staging when holding a caseworker role the same as they would be for a non-user manager caseworker in prod for QA and debugging.

## Notes for reviewer
Please can someone read the performance_tracking/viewing_spec fairly carefully to check my logic.

My assumption has been to that if manually testing a caseworker-only feature on staging a dev will still be set as a user_manager and will want to see what a caseworker would see in prod. And if manually testing a supervisor feature, then would want to see what a supervisor can see and therefore will need to get another user_manager to toggle them to be a supervisor - and only then see the feature.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -51,15 +51,6 @@ class User < ApplicationRecord
     email
   end
 
-  def service_user_supervisor?
-    FeatureFlags.basic_user_roles.enabled? && supervisor? && !can_manage_others?
-  end
-
-  # A dev only user state for debugging in staging, not found in production environments
-  def admin_user_supervisor?
-    FeatureFlags.basic_user_roles.enabled? && supervisor? && can_manage_others?
-  end
-
   def history
     @history ||= AuthorisationHistory.new(user_id: id)
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -51,6 +51,15 @@ class User < ApplicationRecord
     email
   end
 
+  def service_user_supervisor?
+    FeatureFlags.basic_user_roles.enabled? && supervisor? && !can_manage_others?
+  end
+
+  # A dev only user state for debugging in staging, not found in production environments
+  def admin_user_supervisor?
+    FeatureFlags.basic_user_roles.enabled? && supervisor? && can_manage_others?
+  end
+
   def history
     @history ||= AuthorisationHistory.new(user_id: id)
   end
--- a/spec/system/header_nav_spec.rb
+++ b/spec/system/header_nav_spec.rb
@@ -81,10 +81,18 @@ RSpec.describe 'Header navigation' do
                                                  }.from('Your list').to('Manage users')
       end
 
-      it 'shows the "Performance tracking" link and it can be followed' do
-        expect { click_link('Performance tracking') }.to change {
-          page.first('.govuk-heading-xl').text
-        }.from('Your list').to('Performance tracking')
+      it 'does not show the "Performance tracking" link' do
+        expect(page).not_to have_link('Performance tracking')
+      end
+
+      context 'when user managers are logged in as supervisors' do
+        let(:current_user_role) { UserRole::SUPERVISOR }
+
+        it 'shows the "Performance tracking" link and it can be followed' do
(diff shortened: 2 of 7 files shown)


timpeat (commented):
I'm wondering if I shouldn't bring forward the per user reports work. Which, I think, will remove quite a bit of the pain. I'll can try to pull together a spike tomorrow.

app/models/user.rb
```diff
@@ -51,6 +51,15 @@ def name_or_email
     email
   end
 
+  def service_user_supervisor?
+    FeatureFlags.basic_user_roles.enabled? && supervisor? && !can_manage_others?
+  end
+
+  # A dev only user state for debugging in staging, not found in production environments
+  def admin_user_supervisor?
+    FeatureFlags.basic_user_roles.enabled? && supervisor? && can_manage_others?
```
hiboabd, line 60: If a supervisor can already see the performance tracking tab on all environments then I don't see the point of the  `can_manage_others` check. By enabling it for user managers on staging we have to be referring to user managers with the caseworker role

spec/system/header_nav_spec.rb
```diff
@@ -81,10 +81,18 @@
...
 
-      it 'shows the "Performance tracking" link and it can be followed' do
-        expect { click_link('Performance tracking') }.to change {
-          page.first('.govuk-heading-xl').text
-        }.from('Your list').to('Performance tracking')
+      it 'does not show the "Performance tracking" link' do
+        expect(page).not_to have_link('Performance tracking')
+      end
+
+      context 'when user managers are logged in as supervisors' do
```
mseedat-moj, line 88: This test makes it sounds like the user manager can change their role perhaps 'when user manager is supervisor' ?

spec/system/performance_tracking/viewing_spec.rb
```diff
@@ -13,16 +13,49 @@
         heading_text = page.first('.govuk-heading-xl').text
         expect(heading_text).to eq('Manage users')
       end
+
+      it 'does not show the performance tracking nav link' do
+        expect(page).not_to have_link('Performance tracking')
+      end
+
+      context 'when user manager role is supervisor but without service access enabled' do
```
mseedat-moj, line 21: I think this test should explicitly say it is testing the feature flag rather general than system behaviour, maybe even explicitly set the feature flag value to `false`

spec/system/performance_tracking/viewing_spec.rb
```diff
@@ -51,26 +84,22 @@
...
-
-      context 'when logged in as a user manager' do
-        let(:current_user_can_manage_others) { true }
-
-        it 'can access "Performance tracking" page as a user manager' do
-          heading_text = page.first('.govuk-heading-xl').text
-          expect(heading_text).to eq('Performance tracking')
-        end
+        expect(heading_text).to eq('Performance tracking')
       end
```
mseedat-moj, line 103: It seems like there is a whole bunch of tests dedicated to testing the same behaviours with/without the feature flag enabled so maybe laying out the tests with a describe block for 'with feature flag' and 'without feature flag' and then a shared example that runs in both with different expected results will make it a bit clearer?

app/models/user.rb
arthurashman: It pairs with the `service_user_manager?` method that checks that the user is not a user manager.

Those who can see the tab are:
Service supervisors
Admin supervisors AND service access turned on

Everyone else cannot see the nav link or tab content

hiboabd: But on staging what would be the point of setting yourself to be a user manager if you're already a supervisor and therefore should be able to see the page?

spec/system/performance_tracking/viewing_spec.rb
arthurashman: I've tidied things up a bit, but left it as When can access vs When cannot access. I'll DM you

arthurashman:
Not the approach we are taking - different user roles will have access to different reports

PR #392, opened 2023-08-10
CRIMAP-554 Rename "performance tracking" "reports"

timpeat:
## Description of change

Rename "performance tracking" "reports"

## Link to relevant ticket
[CRIMAP-554](https://dsdmoj.atlassian.net/browse/CRIMAP-554)

## Notes for reviewer
This PR paves the way for users who cannot access the casework area of the site (e.g. Data analysts) to view specific reports.

It removes the performance tracking page and, instead, uses the existing reports controller index to list all reports available to a user based on their role.

For the time being, the reporting dashboard is concealed from caseworkers, even though they can still reach their reports via the show action (which is used to display mini reports via turbo on the open and closed applications page).

The vision is that the reporting dashboard will provide users with an index of their reports and access previous monthly and weekly caseworker reports, returns reports, etc.

This PR also removes the Table class which was not well suited to the new forms of report we need to handle.

### Before changes:
Screenshot 2023-08-16 at 16 42 08
Screenshot 2023-08-16 at 16 42 02

### Caseworker
Screenshot 2023-08-16 at 16 39 46

### Supervisor
Screenshot 2023-08-16 at 16 39 28
Screenshot 2023-08-16 at 16 39 19

[CRIMAP-554]: https://dsdmoj.atlassian.net/browse/CRIMAP-554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/controllers/reports_controller.rb
```diff
@@ -1,12 +1,18 @@
...
     case params[:id]
     when /workload_report/
       @report = Reporting::WorkloadReport.new
     when /processed_report/
       @report = Reporting::ProcessedReport.new
-    else
-      raise ActiveRecord::RecordNotFound, 'Report not found'
+    when /caseworker_report/
+      @report = Reporting::CaseworkerReport.new
     end
```
mseedat-moj, line 16: Definitely no need for the `else` ?

app/lib/types.rb
```diff
@@ -53,6 +53,13 @@ module Types
     details: String
   )
 
+  Report = String.enum('caseworker_report', 'processed_report', 'workload_report')
+
+  USER_ROLE_REPORTS = {
+    CASEWORKER_ROLE => [Report['workload_report'], Report['processed_report']],
+    SUPERVISOR_ROLE => Report.values
+  }.freeze
```
mseedat-moj, line 61: Rather than `USER_ROLE_REPORTS` are these `REPORTING_ROLES` ?

app/models/concerns/user_reports.rb
```diff
@@ -0,0 +1,20 @@
+module UserReports
```
mseedat-moj, line 1: Not sure about the module name here - it seems like its about the ability to manager reports about users in my mind! I assume this module is to determine access to Reporting features - can it go in the existing `UserRole` module rather than a separate place?

app/models/concerns/user_reports.rb
```diff
@@ -0,0 +1,20 @@
...
+  extend ActiveSupport::Concern
+
+  def reports
+    return Types::Report.values if user_manager? && FeatureFlags.allow_user_managers_service_access.enabled?
+    return [] if user_manager?
+
+    Types::USER_ROLE_REPORTS.fetch(role)
+  end
```
mseedat-moj, line 9: At the moment if `.fetch` fails because of some dodgy role the exception will be an obtuse error about invalid hash or something. Should there be an explicit check beforehand e.g. `raise LaaException("Unsupported role: '#{role}'") unless Types::USER_ROLE_REPORTS.key?(role)` - i suppose thats an issue for all role access code!

app/models/reporting/processed_report.rb
```diff
@@ -6,8 +6,8 @@ def initialize(day_zero: Time.current)
       @day_zero = day_zero.in_time_zone('London').to_date
     end
 
-    def table
-      Table.new({ processed_on:, applications_closed: }, numeric_column_keys: [])
+    def rows
+      [processed_on, applications_closed].transpose
```
mseedat-moj, line 10: nice. never quite understood transpose except as a pivottable type function!

app/models/reporting/processed_report.rb
```diff
@@ -18,7 +18,7 @@ def applications_closed
       Array.new(3) do |days_ago|
         date_from = day_zero - days_ago
         date_to = date_from.tomorrow
-        Cell.new(closing_events.between(date_from...date_to).count, numeric: false)
+        Cell.new(closing_events.between(date_from...date_to).count, numeric: true)
```
mseedat-moj, line 21: 👍

mseedat-moj (commented):
Looks great. some comments/questions but nothing major to prevent merging tbh.

app/controllers/reports_controller.rb
timpeat: You are right. The controller might not support all user_reports.... Thanks!

app/models/concerns/user_reports.rb
timpeat: I've updated the hash so that it uses the types for the keys, which should hopefully mean that this cannot happen. (I was using constants before, so unlikely, but I think better via the type.)

```
  USER_ROLE_REPORTS = {
    UserRole[CASEWORKER_ROLE] => [Report['workload_report'], Report['processed_report']],
    UserRole[SUPERVISOR_ROLE] => Report.values
  }.freeze
```

app/lib/types.rb
timpeat: These are the reports available to each user depending on their role. Or the user_role's reports.

PR #393, opened 2023-08-11
Tweak docker-compose

zheileman:
## Description of change
With the objective of being able to run all 3 services (apply, review and datastore) at the same time as docker containers, some tweaks are needed.

Prometheus exporter port was hardcoded, it needs to be different in each container otherwise it will clash. Also by default now is disabled as most of the time we will never need it locally, but can be enabled for quick tests.

Added the datastore dummy credentials and endpoint (it works wether datastore is running also in a container, or on the host machine, assuming it's running in port 3003).

Added local docker to SNS bypass, for when Review is running as docker container.

`docker-compose up` will no longer run prometheus exporter. If datastore is up and running in port 3003, Review container will be able to talk to datastore too. SNS will work but requires manually confirming the subscription by visiting the URL logged to stdout.

### DIFF
--- a/app/lib/prometheus_metrics/configuration.rb
+++ b/app/lib/prometheus_metrics/configuration.rb
@@ -4,9 +4,6 @@ module PrometheusMetrics
     require_relative 'collectors'
 
     DEFAULT_PREFIX = 'ruby_'.freeze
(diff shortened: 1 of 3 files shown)


timpeat (approved):
👍

PR #394, opened 2023-08-11
Docker compose works with apple silicon

timpeat:
## Description of change

Add gcombat and update dart-sass to allow ```docker-compose up``` with Apple Silicon

## Notes for reviewer

gcombat allows glibc programs to be run in Alpine Linux (https://wiki.alpinelinux.org/wiki/Running_glibc_programs)

It was also necessary to update dart-sass to a version that correctly understands the architecture it is working in.

To make sure this does not introduce a regression on Intel Mac, remove any docker files you may already have and run  ```docker-compose up```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,8 @@ RUN apk --no-cache add --virtual build-deps \
   linux-headers \
   xz-libs \
   tzdata \
-  yarn
+  yarn \
+  gcompat
 
 # Alpine does not have a glibc, and this is needed for dart-sass
 # Refer to: https://github.com/sgerrand/alpine-pkg-glibc
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,14 +16,6 @@ RUN apk --no-cache add --virtual build-deps \
   yarn \
   gcompat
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -13,7 +13,8 @@ RUN apk --no-cache add --virtual build-deps \
   linux-headers \
   xz-libs \
   tzdata \
-  yarn
+  yarn \
+  gcompat
 
 # Alpine does not have a glibc, and this is needed for dart-sass
```
zheileman, line 19: Maybe this can now be also removed assuming `gcompat` is adding the glibc compatibility?

I can try to test in a fresh M1 (my personal laptop).

zheileman: In my non-M1 machine, it keeps working removing all these lines.
Maybe quickest for you to test in your M1 @timpeat ?

timpeat: Perfect! Will test and update!

timpeat: M1 still requires gcombat sadly. You are correct on the dart-sass fix though.

zheileman (approved):
Great to see that manually installed glibc gone finally! 🥳

PR #399, opened 2023-08-23
Update copy and layout for codefendants

arthurashman:
## Description of change
Updates codefendant table format and copy. Explicitly says if there is a conflict or not in the caption text, and adds badge in No conflict situations
## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-561
### Before changes:
Screenshot 2023-08-21 at 15 12 54

### After changes:
image

Create a case with two codefendants, one with a conflict and one with no conflict.
Check the table looks right. Create a case with no codefendants, make sure table remains how it always has been


willmcb (approved):
👍 tested locally and it works as expected

PR #400, opened 2023-08-23
[CRIMAP-563] Add Data Analyst Role

mseedat-moj:
## Description of change
Introduces a new Data Analyst role 

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-563

## Notes for reviewer
This ticket will require QA as it may have altered expected feature access rights for existing supervisor/caseworkers - specs suggest not.

There is a new routing table in `application_controller` that routes users after sign_in based on their role, rather than defaulting to the 'manage users' page.

Similarly, a new `ForbiddenError` has been introduced which takes the place of the default routing to the 'Manage Users' page.

The primary navigation showing open applications, all applications, etc links has been restricted to 'service users' only.

### Before changes:
N/A

#### New ForbiddenError results in access restricted message
Screenshot 2023-09-08 at 08 24 31

#### Down migration (rollback) is no longer possible
Screenshot 2023-09-08 at 00 25 12

1. Login as a caseworker - should should show all sections as expected
2. Login as supervisor - should show all sections as expected


spec/requests/authentication_spec.rb
```diff
@@ -50,13 +50,13 @@
     end
 
     describe 'when the user is pending authentication' do
-      let(:user) { User.create(email: 'Ben.EXAMPLE@example.com') }
+      let(:user) { User.create(email: 'Ben.EXAMPLE@example.com', role: UserRole::CASEWORKER) }
```
hiboabd, line 53: Is this necessary? Caseworker is the default role

app/controllers/application_controller.rb
```diff
@@ -1,6 +1,22 @@
...
+    render status: :forbidden, template: 'errors/forbidden'
+  end
+
+  def landing_page_for(user)
+    if user.user_manager?
+      admin_manage_users_root_path
+    elsif user.service_user?
+      assigned_applications_path
+    elsif user.reporting_user?
+      reports_path
```
hiboabd, line 16: Supervisors are both service and reporting users. This would route them to the assigned applications page, just confirming that is correct or do we want them to land on the reports page?

app/models/concerns/user_role.rb
```diff
@@ -42,6 +43,10 @@ def service_user?
     [CASEWORKER, SUPERVISOR].include?(role) && !can_manage_others?
   end
 
+  def reporting_user?
+    [SUPERVISOR, DATA_ANALYST].include?(role) && !can_manage_others?
```
arthurashman, line 47: I'm pretty sure a supervisor should also be able to access the service. Just the data analyst that can't - needs double checking

app/controllers/reports_controller.rb
```diff
@@ -17,4 +18,16 @@ def show
...
+
+  def require_dashboard_access!
+    return if current_user.can_access_reporting_dashboard?
+
+    raise ForbiddenError, 'Must be a reporting user'
+  end
+
+  def require_report_access!
+    raise ForbiddenError, 'Report access denied' unless current_user.reports.include?(params[:id])
+  end
```
arthurashman, line 32: In this scenario, I don't think the user doesn't have access, it's that they don't have a report to view. Tim's rework of this made it so that Caseworkers could see their own Caseworker report

app/controllers/reports_controller.rb
```diff
@@ -1,11 +1,12 @@
...
-    raise ActiveRecord::RecordNotFound, 'Reports not found' unless current_user.can_access_reporting_dashboard?
-  end
+class ReportsController < ApplicationController
+  before_action :authenticate_user!
+  before_action :require_dashboard_access!, only: [:index]
+  before_action :require_report_access!, only: [:show]
+  before_action :set_security_headers
 
-  def show
-    raise ActiveRecord::RecordNotFound, 'Report not found' unless current_user.reports.include?(params[:id])
```
arthurashman, line 7: I think this is right

app/models/user.rb
```diff
@@ -18,8 +18,11 @@ class CannotChangeRole < StandardError; end
 
   before_create :set_invitation_expires_at
 
+  # Allow data_analyst users to be deleted because they cannot
+  # be assigned another 'lesser' role - see AddDataAnalystRoleEnum
+  # migration which destroys data_analysts if DB is rolled back
   before_destroy do
-    raise CannotDestroyIfActive if activated?
+    raise CannotDestroyIfActive if activated? && !data_analyst?
```
timpeat, line 25: @mseedat-moj I don't think we want active data analysts to be deletable.

db/migrate/20230823083547_add_data_analyst_role_enum.rb
```diff
@@ -0,0 +1,21 @@
+class AddDataAnalystRoleEnum < ActiveRecord::Migration[7.0]
+  def up
+    execute <<-SQL
+      ALTER TYPE user_role ADD VALUE IF NOT EXISTS '#{Types::DATA_ANALYST_ROLE}';
+    SQL
+  end
+
+  def down
```
timpeat, line 8: I'd rather this down migration raise an error if there are data analysts present in the database.

timpeat (changes requested):
Looks great! Two aspects I'd like to understand better or have changed:
1. remove the code permitting active data analysts to be deleted, 
2. make the down migration raise an error if data analyst present rather than automatically deleting them

app/controllers/reports_controller.rb
mseedat-moj: I checked `main` and this implementation mimics existing behaviour, the difference is the 'access restricted' message is shown rathe than a 'page not found'. Is that incorrect?

db/migrate/20230823083547_add_data_analyst_role_enum.rb
mseedat-moj: Changed to throw an exception on rollback making rollback impossible.

app/models/user.rb
mseedat-moj: This was done to allow rollback, but rollback is now impossible so reverted this rule back

app/models/user.rb
```diff
@@ -18,6 +18,9 @@ class CannotChangeRole < StandardError; end
 
   before_create :set_invitation_expires_at
 
+  # Allow data_analyst users to be deleted because they cannot
```
timpeat, line 21: @mseedat-moj out-dated comment

db/migrate/20230823083547_add_data_analyst_role_enum.rb
timpeat: Thank you :)

timpeat (approved):
Looks great. One comment needs removing, but otherwise good to go!

PR #401, opened 2023-08-25
[CRIMAP-572] Allow user manager to change any role

mseedat-moj:
## Description of change
Instead of the `change role` feature toggling between 'supervisor' and 'caseworker', the User Manager is able to select any other role

DO NOT MERGE UNTIL CRIMAP-563 has been tested and merged!

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-572

## Notes for reviewer
There is no concept of 'lesser' roles so a caseworker can be switched to supervisor and then data analyst. Requires further discovery work.

### Before changes:
Screenshot 2023-08-25 at 09 22 50

### After changes:
Screenshot 2023-08-25 at 09 20 47


db/migrate/20230823083547_add_data_analyst_role_enum.rb
```diff
@@ -0,0 +1,21 @@
...
+  def up
+    execute <<-SQL
+      ALTER TYPE user_role ADD VALUE IF NOT EXISTS '#{Types::DATA_ANALYST_ROLE}';
+    SQL
+  end
+
+  def down
+    data_analysts = User.unscoped.where(role: Types::DATA_ANALYST_ROLE)
+    puts "#{data_analysts.size} Data Analyst(s) will be destroyed!"
+    data_analysts.destroy_all
```
willmcb, line 11: just wondering with this what is the implication if someone moves from caseworker to data analyst and they have done "work" on the system and then is able to be destroyed. Not sure it should matter as I think it is only reference to the ids of users that are used in events.

willmcb (approved):
Tested locally and it worked as expected.

Would like to talk to you about my comment next week though.

db/migrate/20230823083547_add_data_analyst_role_enum.rb
mseedat-moj: Comment no longer applicable because decided to just not allow rollbacks from this day on.

PR #407, opened 2023-09-06
Update deploy workflow with fix for ECR_URL

zheileman:
## Description of change
Same as we did in Datastore and Apply repos.

The previously used `secrets.ECR_URL` is now removed as part of ECR migration to short-term creds so is not available anymore to use in the ./.github/actions/deploy action.

However we can reference job outputs in other job steps to propagate this value.


timpeat (approved):
Thank YOU!

hiboabd (approved):

PR #408, opened 2023-09-08
CRIMAP-586 namespace reporting

timpeat:
## Description of change

Namespace reporting

Also reverts recent changes so as to return "Not found" when users try to access reports they cannot.

Introduces a base ReportingController and namespaces reporting.

## Link to relevant ticket
[CRIMAP-586](https://dsdmoj.atlassian.net/browse/CRIMAP-586)

[CRIMAP-586]: https://dsdmoj.atlassian.net/browse/CRIMAP-586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/controllers/application_controller.rb
```diff
@@ -3,17 +3,13 @@ class ApplicationController < ActionController::Base
 
   class ForbiddenError < StandardError; end
 
-  rescue_from ForbiddenError do |_e|
-    render status: :forbidden, template: 'errors/forbidden'
-  end
-
```
timpeat, line 9: Have shifted this to config/applications.rb

app/controllers/reporting/reports_controller.rb
```diff
@@ -0,0 +1,27 @@
+module Reporting
+  class ReportsController < ReportingController
```
willmcb, line 2: What is `Reporting` in this context?

So are we going to have things other than Reports inherting from the Reporting Controller?

If so I can understand the idea of having "reporting" as a more abstract idea in the application. 

But I think this is a little confusing as well.

If we're not expecting anything other than reports to inherit from this - perhaps we could go with `BaseReportsController` 

Still makes it hard to understand why `ReportsController` would inherit from it though as that too seem like a generic name... 🤔

app/controllers/reporting/reports_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    def show; end
+
+    private
+
+    def set_report
+      @report = report_type_report_mapping[@report_type]
+    end
+
+    def report_type_report_mapping
```
willmcb, line 19: Why not just specify the class directly as before - what is the benefit of the hash lookup - you still have to pass a string specifying the report in the CrimeApplicationController so are we not just doing the same thing with more steps?

app/views/crime_applications/index.html.erb
```diff
@@ -33,7 +33,7 @@
         </span>
       </summary>
       <div class="govuk-details__text ">
-        <%= turbo_frame_tag @report.id, src: report_path(@report.id), data: { turbo: 'true' }, loading: :lazy %>
+        <%= turbo_frame_tag @report_type, src: reporting_report_path(@report_type), data: { turbo: 'true' }, loading: :lazy %>
```
willmcb, line 36: Will this work if we JS is switched off?

spec/system/accessibility_spec.rb
```diff
@@ -88,12 +88,12 @@
     end
 
     it 'workload report page has no axe detectible accessibility issues' do
-      visit report_path(:workload_report)
+      visit reporting_report_path(:workload_report)
       expect(page).to be_axe_clean.according_to accessibility_standards
     end
 
     it 'processed report page has no axe detectible accessibility issues' do
```
willmcb, line 95: I know this is not you but given we are here in the file can we change detectible --> detectable :)

willmcb (commented):
Would be good to chat this through 👍

willmcb (approved):
👍

PR #412, opened 2023-09-14
CRIMAP-543 view latest full week

timpeat:
## Description of change

View full weeks report.

## Link to relevant ticket
[CRIMAP-543](https://dsdmoj.atlassian.net/browse/CRIMAP-543)

[CRIMAP-543]: https://dsdmoj.atlassian.net/browse/CRIMAP-543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/reporting/weekly_report.rb
+++ b/app/models/reporting/weekly_report.rb
@@ -1,12 +1,12 @@
 module Reporting
-  class MonthlyReport
+  class WeeklyReport
     include Reporting::Reportable
 
-    PARAM_FORMAT = '%Y-%B'.freeze
+    PARAM_FORMAT = '%G-%V'.freeze
 
     def initialize(date:, report_type:)
       @date = date
-      @report_type = Types::MonthlyReportType[report_type]
+      @report_type = Types::WeeklyReportType[report_type]
     end
 
     attr_reader :date, :report_type
@@ -15,20 +15,20 @@ module Reporting
       date.strftime(PARAM_FORMAT)
     end
 
-    def month_name
-      date.strftime('%B, %Y')
+    def week_name
+      date.strftime('Week %V, %G')
     end
 
     def title
-      report_text(:title, month_name:)
+      report_text(:title, week_name:)
     end
 
     def start_on
-      date.beginning_of_month
+      date.beginning_of_week
     end
 
     def end_on
-      date.end_of_month
+      date.end_of_week
     end
 
     private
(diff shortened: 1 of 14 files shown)


config/locales/en.yml
```diff
@@ -3,6 +3,7 @@ en:
     DATE_FORMATS: &DATE_FORMATS
       default: '%-d %b %Y'  # 3 Jan 2019
       compact: '%d/%m/%Y'   # 03/01/2019
+      long: '%a %-d %b %y'   # Fri 1 Sep 2023
```
hiboabd, line 6: Not important but i would've preferred it to use the same dates as lines 4 and 5 so `Thur 3 Jan 2019`

spec/system/reporting/weekly_reports_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      last_name: 'Smitheg',
+      auth_oid: '976658f9-f3d5-49ec-b0a9-485ff8b308fa',
+      email: 'Fred.Smitheg@justice.gov.uk'
+    )
+
+    Assigning::AssignToUser.new(
+      user_id: user.id,
+      to_whom_id: user.id,
+      assignment_id: crime_application_id
+    ).call
```
hiboabd, line 22: This feels like it ought to go in a shared context

app/models/reporting/weekly_report.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    def end_on
+      date.end_of_week
+    end
+
+    private
+
+    def stream_name
+      return nil unless report_type == 'caseworker_report'
```
hiboabd, line 37: Could we avoid hardcoding `caseworker_report`

app/views/reporting/user_reports/index.html.erb
```diff
@@ -22,6 +22,17 @@
       </h3>
     <% end %>
 
+    <h2 class='govuk-heading-m'>Latest complete weekly reports<h2>
```
hiboabd, line 25: Do you want to use an `i18n` here?

hiboabd (approved):
LGTM just small picky comments 😁

app/views/reporting/user_reports/index.html.erb
timpeat: This page is very much a test landing page so that Analysts / testers can get to the report under development. I imagine it will be very different once interaction designers have looked at it.

app/models/reporting/weekly_report.rb
timpeat: I've removed this guard as it is no longer needed.

PR #416, opened 2023-09-18
CRIMAP-588 daily reports

timpeat:
## Description of change
Add daily reports.
Refactor monthly, daily and weekly reports to use TemporalReport base class.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-588


app/models/reporting/monthly_report.rb
```diff
@@ -1,43 +1,30 @@
 module Reporting
-  class MonthlyReport
+  class MonthlyReport < TemporalReport
     PARAM_FORMAT = '%Y-%B'.freeze
+    INTERVAL = Types::TemporalInterval['week']
```
arthurashman, line 4: Should this be 'month'?

mseedat-moj (approved):
looks great -some minor suggestions

app/controllers/reporting/temporal_reports_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    def temporal_report_klass
+      case @interval
+      when 'day'
+        Reporting::DailyReport
+      when 'week'
+        Reporting::WeeklyReport
+      when 'month'
+        Reporting::MonthlyReport
+      end
```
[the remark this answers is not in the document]
timpeat: Good idea. I was using the week, month, day to dynamically do time calculations but I'm not doing that now.

app/lib/types.rb
```diff
@@ -56,7 +56,8 @@ module Types
   )
 
   Report = String.enum('caseworker_report', 'processed_report', 'workload_report')
-  WeeklyReportType = MonthlyReportType = String.enum(Report['caseworker_report'])
+  TemporalReportType = WeeklyReportType = MonthlyReportType = String.enum(Report['caseworker_report'])
```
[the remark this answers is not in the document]
timpeat: Good spot, no longer required for refactor.:)

app/models/reporting/monthly_report.rb
timpeat: yes! Thank you.

app/controllers/reporting/temporal_reports_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+      @report = Reporting.const_get(Types::TemporalInterval[@interval].camelize).from_param(
+        report_type: @report_type, period: params[:period]
+      )
+    end
+
+    private
+
+    def set_interval
+      @interval = params.require(:interval).presence_in(
+        *Types::TemporalInterval
```
willmcb, line 17: For the record I like Temporal

app/lib/types.rb
```diff
@@ -56,7 +56,8 @@ module Types
   )
 
   Report = String.enum('caseworker_report', 'processed_report', 'workload_report')
-  WeeklyReportType = MonthlyReportType = String.enum(Report['caseworker_report'])
+  TemporalReportType = String.enum(Report['caseworker_report'])
```
willmcb, line 59: Why not just pass the string?

app/models/reporting/temporal_report.rb
```diff
@@ -0,0 +1,99 @@
...
+    include ActiveModel::Model
+
+    attr_reader :date, :report_type
+
+    def initialize(date:, report_type:)
+      @date = date
+      @report_type = Types::TemporalReportType[report_type]
+    end
+
+    def to_param
```
willmcb, line 12: Might be worth a comment here to explain that PARAM_FORMAT, and constants like these are defined in subclasses.

willmcb: Perhaps you could have:

```ruby
def to_param
  date.strftime(param_format)
end

def param_format
  raise 'Implement in subclass.'
end
```

And then in subclasses:
def param_format
  PARAM_FORMAT
end
```

That way the base has to know less about the concrete implementation in the subclass

File: app/views/reporting/user_reports/index.html.erb [hunk 7]
```diff
@@ -10,24 +10,45 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
-    <h2 class='govuk-heading-m'>Latest complete monthly reports<h2>
+    <h2 class='govuk-heading-m'>Latest complete reports<h2>
```
willmcb, line 13: Are we using i18n for these or are you waiting until the copy / positioning is settled?

app/views/reporting/user_reports/index.html.erb
```diff
@@ -10,24 +10,45 @@
...
     <% end %>
 
-    <h2 class='govuk-heading-m'>Latest complete weekly reports<h2>
     <% @weekly_reports.each do |report|%>
       <h3 class='govuk-heading-s'>
         <%= govuk_link_to(
               report.title,
-              reporting_weekly_report_path(report_type: report.report_type, epoch: report.to_param),
+              reporting_temporal_report_path(
+                interval: 'weekly',
```
willmcb, line 34: Small thing but this could probably be a partial rather than two similar groups of code.

config/brakeman.ignore
```diff
@@ -3,110 +3,76 @@
...
-        "type": "template",
-        "template": "reporting/weekly_reports/show"
-      },
-      "user_input": "params.require(:report_type).presence_in(*Types::Report)",
-      "confidence": "High",
-      "cwe_id": [
-        22
-      ],
-      "note": "Render path restricted by Types::Report."
+      "note": ""
```
willmcb, line 35: Could you add the explanation back in if we are still excusing this please 👍

config/brakeman.ignore
```diff
@@ -3,110 +3,76 @@
...
       },
-      "user_input": "params.require(:report_type).presence_in(*Types::Report).camelize",
-      "confidence": "High",
+      "user_input": "params.require(:interval).presence_in(*Types::TemporalInterval)",
+      "confidence": "Medium",
       "cwe_id": [
         470
       ],
-      "note": "Constant name restricted by Types:Report"
+      "note": ""
```
willmcb, line 58: Same here please :)

config/brakeman.ignore
```diff
@@ -123,9 +89,32 @@
...
+        "type": "method",
+        "class": "Reporting::UserReportsController",
+        "method": "show"
+      },
+      "user_input": "params.require(:report_type).presence_in(*Types::Report)",
+      "confidence": "Medium",
+      "cwe_id": [
+        470
+      ],
+      "note": ""
```
willmcb, line 115: Another explainer here as well please 👍

spec/system/reporting/weekly_reports_spec.rb
```diff
@@ -48,17 +22,17 @@
 
   it 'includes a link to the next weeks\'s report' do
```
willmcb, line 23: Just noted this "weeks's" would be good to correct 👍

willmcb (commented):
Looks good - I've added a few comments we could chat through if you'd like :)

app/views/reporting/user_reports/index.html.erb
timpeat: yes, I imagine this page will change radically once designed :)

PR #421, opened 2023-09-26
CRIMAP-597 Show passporting benefit type on review page

arthurashman:
## Description of change
Surface passporting benefit type when it is provided.

If the applicant is passported but it is a historic application that did not have benefit type requested, list as 'Not requested'

If under 18, (or error has let through a non-means passported applicant) whole row is not shown. 

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-597

### Before changes:
image

### After changes:
image

### DIFF
--- a/app/models/crime_application.rb
+++ b/app/models/crime_application.rb
@@ -27,14 +27,14 @@ class CrimeApplication < LaaCrimeSchemas::Structs::CrimeApplication
     ].join ' '
   end
 
-  delegate :date_of_birth, to: :applicant, prefix: true
+  delegate :date_of_birth, :benefit_type, to: :applicant, prefix: true
 
   def means_passported?
     !means_passport.empty?
   end
 
   def passporting_benefit?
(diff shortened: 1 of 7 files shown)


app/models/crime_application.rb
```diff
@@ -33,6 +33,12 @@ def means_passported?
     !means_passport.empty?
   end
 
+  def passporting_benefit?
+    means_passport.include?('on_benefit_check')
+  end
+
+  delegate :benefit_type, to: :applicant
```
zheileman, line 40: There is already a `delegate :date_of_birth`, maybe worth adding `benefit_type` to that same delegate. It uses prefix tho, so it will become `applicant_benefit_type`, unsure if that causes any trouble, seems fine to me. Just a suggestion 👍

Another suggestion if possible is to use the schemas type instead of a string here `means_passport.include?('on_benefit_check')`

timpeat (approved):
Looks great. Thank you!

PR #422, opened 2023-09-26
Only show one ioj passport, with under 18 taking precedence

arthurashman:
## Description of change
Review should only show one IoJ passport
If there are both IoJ passports present ( 'on_offence', and 'on_age_under18') age passport is the only one that should be shown.

### DIFF
--- a/app/models/crime_application.rb
+++ b/app/models/crime_application.rb
@@ -37,7 +37,14 @@ class CrimeApplication < LaaCrimeSchemas::Structs::CrimeApplication
     means_passport.include?(Types::MeansPassportType['on_benefit_check'])
   end
 
-  delegate :benefit_type, to: :applicant
+  def relevant_ioj_passport
+    # Being under 18 trumps any other interest of justice
+    if ioj_passport.include?('on_age_under18')
+      'on_age_under18'
+    elsif ioj_passport == ['on_offence']
+      'on_offence'
+    end
+  end
 
   def history
     @history ||= ApplicationHistory.new(application: self)
--- a/app/views/casework/crime_applications/_interests_of_justice.html.erb
+++ b/app/views/casework/crime_applications/_interests_of_justice.html.erb
@@ -27,17 +27,6 @@
       <% end %>
     </tbody>
   </table>
-<% elsif crime_application.ioj_passport.include?('on_age_under18') %>
-  <dl class="govuk-summary-list govuk-!-margin-bottom-9">
-    <div class="govuk-summary-list__row">
-      <dt class="govuk-summary-list__key">
-        <%= label_text(:justification) %>
-      </dt>
-      <dd class="govuk-summary-list__value">
-        <%= t('on_age_under18', scope: 'values') %>
-      </dd>
-    </div>
-  </dl>
 <% else %>
   <dl class="govuk-summary-list govuk-!-margin-bottom-9">
     <div class="govuk-summary-list__row">
@@ -45,7 +34,11 @@
         <%= label_text(:justification) %>
       </dt>
       <dd class="govuk-summary-list__value">
-        <%= t('on_offence', scope: 'values') %>
+        <% if crime_application.ioj_passport.include?('on_age_under18') %>
+          <%= t('on_age_under18', scope: 'values') %>
+        <% else %>
+          <%= t(crime_application.ioj_passport.first, scope: 'values') %>
+        <% end %>
       </dd>
     </div>
   </dl>
--- a/app/views/casework/crime_applications/_interests_of_justice.html.erb
+++ b/app/views/casework/crime_applications/_interests_of_justice.html.erb
@@ -34,11 +34,7 @@
         <%= label_text(:justification) %>
       </dt>
       <dd class="govuk-summary-list__value">
-        <% if crime_application.ioj_passport.include?('on_age_under18') %>
-          <%= t('on_age_under18', scope: 'values') %>
-        <% else %>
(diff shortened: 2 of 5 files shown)


app/views/casework/crime_applications/_interests_of_justice.html.erb
```diff
@@ -27,17 +27,26 @@
...
-          <%= t(passport, scope: 'values') %>
-        </dd>
-      </div>
-    <% end %>
+    <div class="govuk-summary-list__row">
+      <dt class="govuk-summary-list__key">
+        <%= label_text(:justification) %>
+      </dt>
+      <dd class="govuk-summary-list__value">
+        <%= t('on_offence', scope: 'values') %>
```
arthurashman, line 48: Given what happened with the means_passport, I wonder if it would be foolhardy to have this as a default else if there is no ioj_reasons, and the ioj_passport does not contain the age passport. 

@timpeat I'd appreciate your thoughts. Is there a sensible default else clause to show them something went wrong as a fallback for if Apply lets something through that it shouldn't

timpeat: I'd be inclined to do something like:

``` ruby
      <dd class="govuk-summary-list__value">
        <%= t(crime_application.ioj_passport.sort.join('_and_'), scope: 'values') %>
        </dd>
```
and then add a translation for key:
``` yaml
    on_age_under18_and_on_offence: Not needed based on offence and because the client is under 18 years old
    on_age_under18: Not needed because the client is under 18 years old
    on_offence: Not needed based on offence

**Review by timpeat — COMMENTED** [review]
See comment for my suggestion of how to make it dynamic. Obviously you don't need to use my suggested text for the "on_age_under18_and_on_offence:" key.

File: app/views/casework/crime_applications/_interests_of_justice.html.erb [hunk 2]
```diff
@@ -29,15 +29,17 @@
...
-          <%= t(passport, scope: 'values') %>
-        </dd>
-      </div>
-    <% end %>
+    <div class="govuk-summary-list__row">
+      <dt class="govuk-summary-list__key">
+        <%= label_text(:justification) %>
+      </dt>
+      <dd class="govuk-summary-list__value">
+        <% if crime_application.ioj_passport.include?('on_age_under18') %>
```
willmcb, line 37: Could have a presenter for this if we think that this logic is a bit much for the views - though happy this way.

app/views/casework/crime_applications/_interests_of_justice.html.erb
timpeat: > So when it is passported on under 18 that trumps the passport on offence, so they want that to be the only one that shows.

In which case you could just change "on_age_under18_and_on_offence:" value in locale file to "Not needed because the client is under 18 years old"

app/views/casework/crime_applications/_interests_of_justice.html.erb
willmcb: Obviously change the method names to make more sense as you like :)

willmcb (changes requested):
Happy to chat through if you like 👍

app/models/crime_application.rb
```diff
@@ -37,7 +37,13 @@ def passporting_benefit?
     means_passport.include?(Types::MeansPassportType['on_benefit_check'])
   end
 
-  delegate :benefit_type, to: :applicant
```
arthurashman, line 40: This was duplicated on line 30 by accident

app/views/casework/crime_applications/_interests_of_justice.html.erb
```diff
@@ -29,15 +29,13 @@
...
-          <%= t(passport, scope: 'values') %>
-        </dd>
-      </div>
-    <% end %>
+    <div class="govuk-summary-list__row">
+      <dt class="govuk-summary-list__key">
+        <%= label_text(:justification) %>
+      </dt>
+      <dd class="govuk-summary-list__value">
+        <%= t(crime_application.precedent_ioj_passport, scope: 'values') %>
```
willmcb, line 37: Is this ment to be `present_ioj_passport` rather than precedent?

app/models/crime_application.rb
```diff
@@ -37,7 +37,13 @@ def passporting_benefit?
...
+  def precedent_ioj_passport
+    if ioj_passport.include?('on_age_under18')
+      'on_age_under18'
+    elsif ioj_passport == ['on_offence']
```
arthurashman, line 43: Apply is now catching if there is neither an Ioj reason, nor ioj passport, so we may not need this to be an explicit elsif, but it felt a bit safer. Although I wasn't sure what default else message to add, so it will blow up if Apply fails somehow

app/views/casework/crime_applications/_interests_of_justice.html.erb
arthurashman: Precedent i.e. the most important - I struggled to find the right name. https://en.wiktionary.org/wiki/precedent#:~:text=justify%20a%20judgment-,Adjective,arrangement%3B%20preceding%2C%20foregoing.%20%5B

Wiktionary say that the usage I have gone for is 'now rare'

arthurashman: Maybe I was thinking https://www.merriam-webster.com/dictionary/preeminent - would that be any better?? Is there anything you'd suggest?

willmcb (approved):
👍

PR #423, opened 2023-09-28
allow_user_managers_service_access to reman true in local

arthurashman:
## Description of change

Modify tests to allow for allow_user_managers_service_access to reman true in local

Toggling allow_user_managers_service_access in local development feels unnecessary and more helpful to leave it on.


timpeat (changes requested):
Use managers having to the service is an exception for the staging environment only. I'd like to keep all other ENVs as closed the the live service as possible. 

It would be good to find out why you'd like to make this change as its never caused me any issues.

PR #424, opened 2023-10-02
Crimap 526 list supporting evidence

MohammedHassanLatif93:
## Description of change
Added the supporting evidence section, added the feature behind an “evidence upload” feature flag. This is part one of two part pr's. In the next pr we will be adding the download functionality.

## Link to relevant ticket
[CRIMAP-526](https://dsdmoj.atlassian.net/browse/CRIMAP-526)
[CRIMAP-536 - Designs can be found here](https://dsdmoj.atlassian.net/browse/CRIMAP-536)

### After changes:
image

You will need to submit an application in Apply with supporting evidence, you can then view the application in review.

[CRIMAP-526]: https://dsdmoj.atlassian.net/browse/CRIMAP-526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/lib/types.rb
```diff
@@ -74,4 +74,19 @@ module Types
     applicant_name
   ].freeze
   SortBy = String.default('submitted_at'.freeze).enum(*SORTABLE_COLUMNS)
+
+  FILE_TYPES = {
```
zheileman, line 78: Might be simplified by doing `File.extname(filename)`?
Just an idea. As it seems the important thing is the extension of the file.

hiboabd: Thanks, i've added that

config/locales/en/casework.yml
```diff
@@ -195,6 +197,7 @@ en:
       sent_back: red
       completed: green
       marked_as_ready: blue
+    download_file: Download file (%{content_type}, %{file_size})
```
timpeat, line 200: Can we get a line break between download file and file size as per design?

app/models/crime_application.rb
```diff
@@ -1,9 +1,14 @@
 require 'laa_crime_schemas'
 
+# rubocop:disable Metrics/ClassLength
 class CrimeApplication < LaaCrimeSchemas::Structs::CrimeApplication
   include Assignable
   include Reviewable
 
+  def file_type(filename)
```
timpeat, line 8: Can this method be moved? Happy suggest an alternative.

timpeat (changes requested):

app/models/document.rb
```diff
@@ -0,0 +1,7 @@
+require 'laa_crime_schemas'
+
+class Document < LaaCrimeSchemas::Structs::Document
+  def file_type
```
zheileman, line 4: If we think this may be useful, we can extract this helper method to the gem Structs so both Apply/Review can use it. But no urgency for now.

app/views/casework/crime_applications/show.html.erb
```diff
@@ -15,6 +15,9 @@
     <%= render partial: 'offences', object: @crime_application.case_details.offences %>
     <%= render partial: 'codefendants', object: @crime_application.case_details.codefendants %>
     <%= render partial: 'interests_of_justice', locals: { crime_application: @crime_application } %>
+    <% if FeatureFlags.evidence_upload.enabled? %>
```
zheileman, line 18: Might not be necessary as the whole partial is conditional on whether `supporting_evidence.present?` which will always be false until some provider start uploading files?

config/locales/en/casework.yml
```diff
@@ -195,6 +197,7 @@ en:
       sent_back: red
       completed: green
       marked_as_ready: blue
+    download_file: "Download file <br> (%{content_type}, %{file_size})"
```
zheileman, line 200: Instead of `content_type` perhaps `file_type` or `file_extension` is more close to what we are doing/presenting? No blocker at all just a suggestion.

app/views/casework/crime_applications/show.html.erb
hiboabd: Yeah agreed, this shouldn't be displayed as it's not being used yet.  But it was part of the ticket to be behind a feature flag so we followed that for now

config/locales/en/casework.yml
hiboabd: Yeah `file_extension` is most accurate probably, will change

app/models/document.rb
hiboabd: Yeah we considered that but thought it would not be required on apply. Depends on if apply will ever present/need the file extension...

hiboabd: I'm not even sure caseworkers need it if it's visible in the file name but 🤷‍♀️

config/settings.yml
```diff
@@ -11,3 +11,7 @@ feature_flags:
     local: true
     staging: true
     production: false
+  evidence_upload:
+    local: true
+    staging: true
+    production: false
```
mseedat-moj, line 17: Needs a new line at the end?

spec/system/casework/viewing_an_application/that_is_open_and_unassigned_spec.rb
```diff
@@ -345,4 +345,14 @@
       expect(page).to have_content('Home address Does not have a home address')
     end
   end
+
+  context 'with supporting evidence' do
+    it 'shows a table with supporting evidence' do
+      evidence_row = find(:xpath,
+                          "//table[@class='govuk-table app-dashboard-table govuk-!-margin-bottom-9']
+                          //tr[contains(td[1], 'test.pdf')]")
```
mseedat-moj, line 353: Is there another way to find this table using xpath (or css)? e.g. by picking out the '3rd table in the page' or something like that?

spec/system/casework/viewing_an_application/that_is_open_and_unassigned_spec.rb
```diff
@@ -345,4 +345,14 @@
...
+
+      expect(evidence_row).to have_content('test.pdf Download file (pdf, 12 Bytes)')
+    end
+  end
```
mseedat-moj, line 357: Does there need to be any negative tests e.g. if the supporting evidence file is no longer present for a revision? Or is that better for a future ticket where multiple submissions with different evidence can be tested out?

app/models/document.rb
```diff
@@ -0,0 +1,7 @@
+require 'laa_crime_schemas'
+
+class Document < LaaCrimeSchemas::Structs::Document
+  def file_extension
+    File.extname(filename).delete_prefix('.')
```
mseedat-moj, line 5: 👍

app/models/crime_application.rb
```diff
@@ -1,9 +1,14 @@
 require 'laa_crime_schemas'
 
+# rubocop:disable Metrics/ClassLength
 class CrimeApplication < LaaCrimeSchemas::Structs::CrimeApplication
   include Assignable
   include Reviewable
 
+  def supporting_evidence
+    super.map { |d| Document.new(d.attributes) }
+  end
```
mseedat-moj, line 10: Showing my ignorance here - but is `super` a whole list of crime applications then?

mseedat-moj (approved):
Looking good don't see any major issues. Left some comments to answer.

app/models/crime_application.rb
hiboabd: A list of supporting evidence - d stands for documents

spec/system/casework/viewing_an_application/that_is_open_and_unassigned_spec.rb
hiboabd: Probably but I'm following the pattern in the file, I think this would need refactoring for the whole file in that case

spec/system/casework/viewing_an_application/that_is_open_and_unassigned_spec.rb
hiboabd: I've added a test to check that if there is no supporting evidence then no supporting evidence section is shown

timpeat (approved):
Looks great!

PR #425, opened 2023-10-03
[CRIMAP-601] Show first court hearing location

mseedat-moj:
## Description of change
Shows the first court hearing location name ('Initial court')

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-601

## Notes for reviewer
Not sure whether the `is_first_court_hearing` value needs to be checked as well to determine output?

### Before changes:
Screenshot 2023-10-03 at 21 53 11

### After changes:
Screenshot 2023-10-03 at 21 50 55

Create and application and in the court hearing section enter appropriate values.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,7 +23,7 @@ gem 'laa-criminal-applications-datastore-api-client',
     require: 'datastore_api'
 
 gem 'laa-criminal-legal-aid-schemas',
-    github: 'ministryofjustice/laa-criminal-legal-aid-schemas', tag: 'v1.0.2'
+    github: 'ministryofjustice/laa-criminal-legal-aid-schemas', tag: 'v1.0.3'
 
 # The original asset pipeline for Rails [https://github.com/rails/sprockets-rails]
 gem 'sprockets-rails'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -9,10 +9,10 @@ GIT
 
 GIT
   remote: https://github.com/ministryofjustice/laa-criminal-legal-aid-schemas.git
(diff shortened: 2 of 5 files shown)


zheileman:
It LGTM. Only thing to consider is if Review should show to the case worker a more dynamic "initial court" row depending on the answer to the "is_first_court_hearing", so:

- When `is_first_court_hearing` is `nil` then we present "This question was not asked when this application was submitted" (check with content, can't remember the exact copy!)

- When `is_first_court_hearing` is `yes` then we present the same value as in the `hearing_court_name` (as in, both are the same)

- When `first_court_hearing_name` is `no` we present the value of `first_court_hearing_name` (because they are different)

- When `first_court_hearing_name` is `no_hearing_yet` hmmm, maybe same copy as in Apply? `There has not been a hearing yet`

All this TBC, not sure what was the original idea behind gathering this data and how to best present to case worker to make an informed decision! Or if it was only about MAAT injection.

willmcb (approved):
Nice one :)

mseedat-moj:
Implemented a complex ruleset and then reverted it until further instruction:

```
<!-- Court hearing -->
<% if is_first_court_hearing == 'no_hearing_yet' %>
  Court hearing the case: There has not been a hearing yet
<% else %>
  Next court hearing date: 12th January 2024

  Court hearing the case: Westminster Magistrates

  <% if is_first_court_hearing == 'no' && first_court_hearing_name.present? %>
    Initial court: Southwark Crown Court
  <% end %>
<% end %>
```

zheileman (approved):
Agreed to introduce a more in-deep logic in a presenter later on once confirmed with Liam.

PR #426, opened 2023-10-04
[CRIMAP-526] download supporting evidence

hiboabd:
## Description of change
This is the second part of two PR's to implement downloading supporting evidence on review. This PR enables users to download evidence. There are two restrictions to downloading evidence: 1) You must be assigned to the application 2) You must be downloading a file that is uploaded to the current application (this is to prevent malicious actors from changing the application id of a link to an application they are assigned to, to download files in other applications). 

Videos are linked below to demonstrate the functionality 

## Link to relevant ticket
[CRIMAP-526](https://dsdmoj.atlassian.net/browse/CRIMAP-526)

## Notes for reviewer
Note there is slightly different download behaviour per browser and we found that with a firefox browser, pdf files are downloaded and also opened in another tab. This didn't occur for png files and we haven't tested all file types so we're not sure what other file types are impacted. Not sure if this is something that can be fixed with JS in some way but we didn't explore for now as it is only happening in one browser and we will potentially explore opening the file in a new tab vs. downloading (according to the ticket). 

Also error message content needs review from content/design.

### After changes:

**Scenario: successfully downloading files**

https://github.com/ministryofjustice/laa-review-criminal-legal-aid/assets/51047911/89aa307f-9234-4047-a1a3-7c582917b599

**Scenario: Attempting to download a file when user is not assigned to the current application**

https://github.com/ministryofjustice/laa-review-criminal-legal-aid/assets/51047911/cbf32ec7-6e73-46e5-b4e2-77f92597a5bb

**Scenario: Attempting to download a file that is not uploaded to the current application**

https://github.com/ministryofjustice/laa-review-criminal-legal-aid/assets/51047911/98e91503-5b8e-4abe-9c66-0ed10276119d

[CRIMAP-526]: https://dsdmoj.atlassian.net/browse/CRIMAP-526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/casework/documents_controller.rb
+++ b/app/controllers/casework/documents_controller.rb
@@ -13,7 +13,7 @@ module Casework
       raise CannotDownloadDocNotPartOfApp unless doc_uploaded_to_current_app?
 
       presign_download = Datastore::Documents::Download.new(document: @document).call
-      redirect_to(presign_download.url, allow_other_host: true)
+      redirect_to(presign_download.url, allow_other_host: true) if presign_download
     rescue CannotDownloadUnlessAssigned
       set_flash(:cannot_download_unless_assigned, success: false)
       redirect_to crime_application_path(params[:crime_application_id])
--- a/app/controllers/casework/documents_controller.rb
+++ b/app/controllers/casework/documents_controller.rb
@@ -3,22 +3,22 @@ module Casework
     before_action :set_crime_application, :set_document
 
     class CannotDownloadUnlessAssigned < StandardError; end
-    class CannotDownload < StandardError; end
+    class CannotDownloadDocNotPartOfApp < StandardError; end
 
     def show; end
 
     # rubocop:disable Metrics/AbcSize
     def download
       raise CannotDownloadUnlessAssigned unless @crime_application.assigned_to?(current_user_id)
-      raise CannotDownload unless doc_uploaded_to_current_app?
+      raise CannotDownloadDocNotPartOfApp unless doc_uploaded_to_current_app?
 
       presign_download = Datastore::Documents::Download.new(document: @document).call
       redirect_to(presign_download.url, allow_other_host: true)
     rescue CannotDownloadUnlessAssigned
       set_flash(:cannot_download_unless_assigned, success: false)
       redirect_to crime_application_path(params[:crime_application_id])
-    rescue CannotDownload
-      set_flash(:cannot_download, success: false)
+    rescue CannotDownloadDocNotPartOfApp
+      set_flash(:cannot_download_doc_uploaded_to_another_app, success: false)
       redirect_to crime_application_path(params[:crime_application_id])
     end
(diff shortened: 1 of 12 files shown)


app/controllers/casework/documents_controller.rb
```diff
@@ -0,0 +1,41 @@
+module Casework
+  class DocumentsController < Casework::BaseController
+    before_action :set_crime_application, :set_document
+
+    class CannotDownloadUnlessAssigned < StandardError; end
+    class CannotDownloadDocNotPartOfApp < StandardError; end
```
hiboabd, line 6: Let me know if this name makes sense? I struggled to keep the name within the rubocop character limit but also convey enough meaning..

zheileman:
One minor suggestion perhaps is to not have a clickable download link if the application is not assigned. Maybe just present size and file type, but without the "Download" word.

app/controllers/casework/documents_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def show; end
+
+    # rubocop:disable Metrics/AbcSize
+    def download
```
timpeat, line 11: It feels a bit odd raising and rescuing in the same method. I wonder if it would be better to have a before filter checks. Something like: "require_assigned_user".

require_assigned_user method would then do check that document is from current_application and current_application is assigned to user.

hiboabd:
> One minor suggestion perhaps is to not have a clickable download link if the application is not assigned. Maybe just present size and file type, but without the "Download" word.

Ooh interesting suggestion, i'll pop in the ucd channel and see what people think

app/controllers/casework/documents_controller.rb
hiboabd: Nice I've made that change

app/controllers/casework/documents_controller.rb
willmcb: I think this is good - it is clear what it is and squeezes a lot of info in to a relitively few characters.

spec/services/datastore/documents/download_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  describe '#call' do
+    context 'when a document download link is retrieved successfully' do
+      before do
+        stub_request(:put, 'https://datastore-api-stub.test/api/v1/documents/presign_download')
+          .with(body: expected_query)
+          .to_return(status: 201, body: { url: presign_download_url }.to_json)
+      end
+
+      it 'returns presign download url' do
```
willmcb, line 18: presign --> presigned

willmcb (approved):
Great work Hibo, happy for this to be merged in.

- I take the points around the copy and the download links, I think we can fix forward as if / when we get answers from UCD

PR #427, opened 2023-10-05
CRIMRE-589 event source the workload mini report

timpeat:
## Description of change
CRIMRE-589 refactor workload report to event source its data.

## Link to relevant ticket
[CRIMAP-589](https://dsdmoj.atlassian.net/browse/CRIMAP-589)

## Notes for reviewer

The workload report has only been available as a snapshot of current state. CRIMRE-589  adds functionality to view this report as it was at any given time. This PR is a prepares for that by event sourcing the report rather than relying on the data cached in the ReceivedOnReport table.

Confirm that the mini report on the open applications page is correct.

[CRIMAP-589]: https://dsdmoj.atlassian.net/browse/CRIMAP-589?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/read_models/received_on_reports/projection.rb
+++ b/app/read_models/received_on_reports/projection.rb
@@ -1,0 +1,43 @@
+module ReceivedOnReports
+  class Projection
+    def initialize(stream_names:, observed_at: Time.current)
+      @stream_names = stream_names.uniq
+      @observed_at = observed_at
+    end
+
+    attr_reader :observed_at
+
+    def dataset
+      return @dataset if @dataset
+
+      total_received = 0
+      total_closed = 0
+
+      @stream_names.each do |stream_name|
+        scope = scope(stream_name)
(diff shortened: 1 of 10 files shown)


willmcb (approved):
Looks great 👍

app/read_models/received_on_reports/projection.rb
```diff
@@ -0,0 +1,43 @@
+module ReceivedOnReports
+  class Projection
```
willmcb, line 2: Discussion - technically this isn't a "projection" could this be a Read Model or somthing else?

timpeat: Good point. I'll consider this when I do the next one and possibly rename this class after that.

PR #432, opened 2023-10-06
CRIMAP-651 Skip attack-sqli check on session cookie

timpeat:
## Description of change

Prevent sql injection false positives on the rails session cookie

## Link to relevant ticket
[CRIMAP-651](https://dsdmoj.atlassian.net/browse/CRIMAP-651)

## Notes for reviewer
False positives on the session cookie have been causing 403 errors. Here we use tags rather than id to prevent the offending rules being applied to the rails session cookie.  After testing on staging, this change will need to be applied to the production ingress.

[CRIMAP-651]: https://dsdmoj.atlassian.net/browse/CRIMAP-651?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/kubernetes/staging/ingress.yml
+++ b/config/kubernetes/staging/ingress.yml
@@ -22,8 +22,9 @@ metadata:
       SecRequestBodyLimit 1048576
       SecRequestBodyNoFilesLimit 1048576
       SecDefaultAction "phase:2,pass,log,tag:github_team=laa-crime-apply"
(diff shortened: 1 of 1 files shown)


zheileman (approved):
🙌

PR #433, opened 2023-10-09
[CRIMAP-654] Add first court hearing only if necessary

mseedat-moj:
## Description of change
Shows first court hearing details based on the `is_first_court_hearing` flag

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-654

## Notes for reviewer
Consider whether presenter is necessary for view logic

### DIFF
--- a/app/presenters/crime_application_presenter.rb
+++ b/app/presenters/crime_application_presenter.rb
@@ -1,0 +1,33 @@
+require 'laa_crime_schemas'
+
+class CrimeApplicationPresenter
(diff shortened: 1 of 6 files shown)


willmcb:
I think on the presenters question there are a few areas where these could be introduced - maybe this is a first place to do so.

Alternatively happy for there to be some tickets on the backlog to add these in - would probably be good tickets to pair on I think.

mseedat-moj:
PR canceled due to changes in the requirements

PR #446, opened 2023-10-30
[CRIMAP-654/662] Introduce presenter pattern using First Court Hearing

mseedat-moj:
## Description of change
Shows "First court hearing" name depending on the `case_details.is_first_court_hearing` and related fields. Does so by introducing the 'Presenter pattern' as found in Apply

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-654 - Show first court hearing
https://dsdmoj.atlassian.net/browse/CRIMAP-662 - Add presenter pattern to Review
https://dsdmoj.atlassian.net/browse/CRIMAP-661 - Touches on 'Add presenter to Case Details'

## Notes for reviewer
- Is the intstantiation of the presenter by overwriting `CrimeApplication#case_details` appropriate?
- Should ` Court Hearing` be a separate partial?

#### 0. Only first court hearing
```
{
  "hearing_court_name": "Westminster", 
  "is_first_court_hearing":"no", 
  "first_court_hearing_name": "Westminster", 
  "hearing_date": "2023-12-01",
}
```
0-first-ever

#### 1. First/Next court same
```
{
  "hearing_court_name": "Westminster", 
  "is_first_court_hearing":"no", 
  "first_court_hearing_name": "Westminster", 
  "hearing_date": "2023-12-01",
}
```
1-not-first-same-court

#### 2. First/Next court not the same
```
{
  "hearing_court_name": "Westminster", 
  "is_first_court_hearing":"no", 
  "first_court_hearing_name": "Cardiff", 
  "hearing_date": "2023-12-01",
}
```
2-not-first-different

#### 3. There has not been a hearing yet
```
{
  "hearing_court_name": "Westminster", 
  "is_first_court_hearing": null, 
  "first_court_hearing_name": "Cardiff", 
  "hearing_date": "2023-12-01",
}
```
Screenshot 2023-10-30 at 12 03 22

#### 4. Legacy application
```
{
  "hearing_court_name": "Westminster", 
  "is_first_court_hearing": null, 
  "first_court_hearing_name": "Cardiff", 
  "hearing_date": "2023-12-01",
}

```
4-legacy


app/decorators/base_decorator.rb
```diff
@@ -0,0 +1,5 @@
+class BaseDecorator < SimpleDelegator
```
willmcb, line 1: Just thinking - are we using decorators at the moment? What will they do differently to presenters?

Thinking we could just have presenters for now, and not introduct a very similar concept without being really clear why.

What do you think @mseedat-moj ? 🤔

mseedat-moj: I brought it in to mimic Apply and set up any subsequent tickets. The `BaseDecorator` and `BasePresenter` were added to Apply at the same time last year. My understanding is the `Presenter` is designed to output HTML along with general UI based methods, and decorators are 'purer' non-output related classes. I think the distinction is academic and I've used a Presenter in this PR for example rather than a Decorator.

I'm inclined to agree that the `BaseDecorator` in Apply and Review are both redundant.

willmcb (approved):

app/decorators/base_decorator.rb
mseedat-moj: Discussion here: https://mojdt.slack.com/archives/C03JS4V9TPU/p1698839205005959

PR #448, opened 2023-11-02
[CRIMAP-661] Introduce ApplicantPresenter

MohammedHassanLatif:
## Description of change
Tidied up code and introduced applicant presenter. Moved applicant related methods from crime_application.rb into applicant_presenter.rb. Removed delegations for date_of_birth and benefit_type as they can be called directly from the applicant presenter.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-661


mseedat-moj (changes requested):
CrimeApplication class has accidentally changed the line numbers for 2 methods so they need to go back to where they originally were.

app/views/casework/crime_applications/show.html.erb
```diff
@@ -9,7 +9,7 @@
                locals: {
                  applicant: @crime_application.client_details.applicant,
                  nino: @crime_application.formatted_applicant_nino,
-                 applicant_tel: @crime_application.formatted_applicant_telephone_number
+                 applicant_tel: @crime_application.phone_number.for_applicant
```
hiboabd, line 12: Feels like the order should read, `crime_application.applicant.phone_number` or `crime_application.applicant.formatted_phone_number` or something. This reads like the crime application has a phone number. 

In that case, would it be better to have an applicant presenter? I see there is also a `formatted_applicant_nino` method on line 11 so looks like there are a few methods that could move into that

timpeat (approved):
Nice one @MohammedHassanLatif

app/models/crime_application.rb
```diff
@@ -126,32 +107,7 @@ def case_details
...
-    when /\A\+4407\d{9}\z/ # +44 mobile number format with initial 0
-      [3, 9, 13].each { |i| formatted_tel.insert i, ' ' }
-    when /\A\+44\(0\)7\d{9}\z/ # +44 mobile numberformat with initial 0 in parentheses
-      [3, 7, 12, 16].each { |i| formatted_tel.insert i, ' ' }
-    else
-      formatted_tel = applicant.telephone_number
-    end
-
-    formatted_tel
+    @applicant ||= ApplicantPresenter.present(self[:client_details][:applicant])
```
mseedat-moj, line 111: Is there ever situation where `{ "client_details": { "applicant": null } }` that is - there is no applicant available?

mseedat-moj (approved):
Well done @MohammedHassanLatif @hiboabd - question from me about handling possible nil?

app/models/crime_application.rb
hiboabd: Nope there should always be an applicant - https://github.com/ministryofjustice/laa-criminal-legal-aid-schemas/blob/main/schemas/1.0/application.json

PR #452, opened 2023-11-09
[CRIMAPP-92] Introduce interface to allocate caseworker to work stream

hiboabd:
## Description of change
Note this PR is the first of a couple of PR's that will enable supervisors to manage which work streams caseworkers have access to and caseworkers to work from a particular work stream. 

This PR introduces the 'Manage Competencies' interface to enable supervisors to view a list of caseworkers and their competencies (currently hardcoded). The next PR will handle the functionality to manage these competencies. 

Also adds a work stream feature flag which the tab is hidden behind. 

## Link to relevant ticket
[CRIMAPP-92](https://dsdmoj.atlassian.net/browse/CRIMAPP-92)

## Notes for reviewer
Does competencies feel right in describing what the workstreams represent in relation to a user? Skills, capabilities were also considered.

I introduced a couple of shared examples refactoring some duplicate spec testing around the pagination and ordering of a table (of users in this case). They could probably be made even more flexible but I'm not really sure if there will be many specs using these 

### After changes:
Screenshot 2023-11-09 at 17 07 12
Screenshot 2023-11-09 at 17 07 35

Requires you to have at least one supervisor and caseworker user. 
- Log into review as a supervisor. You should see the manage competencies tab and be able to click onto it to view the dashboard. There you should be able to see a list any active caseworkers with a hardcoded competency list
- Log into review as a non supervisor role, you should not be able to view the manage competencies tab and if you navigate to `/manage_competencies`, you should see an error 
- You can also test feature flag functionality by changing your local `settings.yml` value to false and rebuilding your local server. Check you can't see the tab when logged in as a supervisor.

[CRIMAPP-92]: https://dsdmoj.atlassian.net/browse/CRIMAPP-92?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/controllers/manage_competencies/base_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+    def text_namespace
+      :manage_competencies
+    end
+
+    private
+
+    def require_supervisor!
+      return if current_user.role == Types::SUPERVISOR_ROLE
+
+      render status: :not_found, template: 'errors/not_found', layout: 'external'
```
timpeat, line 19: Doing this:
```
      raise ForbiddenError, 'Must be a supervisor'
```
will have the same effect

timpeat (approved):
One minor, non-blocking, comment.

app/controllers/manage_competencies/base_controller.rb
hiboabd: Thanks, made that change

PR #456, opened 2023-11-14
[CRIMAPP-91] display caseworker competency history

hiboabd:
## Description of change
Adds a history link to the manage competency page to enable supervisors to view the history of changes to a caseworkers competencies. If there is no history, the table is not visible and placeholder text is displayed. 

Also moves the link to change competencies from the caseworkers name to the competency list (see screenshots below).

## Link to relevant ticket
[CRIMAPP-91](https://dsdmoj.atlassian.net/browse/CRIMAPP-91)

### Before changes:
Screenshot 2023-11-14 at 14 21 08

### After changes:
Screenshot 2023-11-14 at 13 58 43
Screenshot 2023-11-14 at 13 31 10

Screenshot 2023-11-14 at 14 12 53
Screenshot 2023-11-14 at 14 13 03

Requires you to have at least one supervisor and caseworker user.

Log into review as a supervisor. You should see the manage competencies tab and be able to click onto it to view the dashboard. 

There you should be click onto the history link that will take you the history page. 

If you have not set any competencies for your caseworker, then you should not see the history table and in it's place you should see the text 'There is no history'.

Play around with setting/unsetting the competencies of your caseworker and confirm that you can see the correct record of changes in the history page. 

[CRIMAPP-91]: https://dsdmoj.atlassian.net/browse/CRIMAPP-91?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/allocation_history.rb
```diff
@@ -0,0 +1,13 @@
...
+  attribute :user_id, Types::Uuid
+
+  def items
+    @items ||= events.map { |event| AllocationHistoryItem.new(event) }
+  end
+
+  private
+
+  def events
+    Allocating.user_events(user_id).backward
```
timpeat, line 11: "backwards" is already applied to this scope. I wonder if it should be removed from "Allocating.user_events" and added by the caller where needed? I.e. Allocating.user_competencies and this method.

app/models/allocation_history_item.rb
```diff
@@ -0,0 +1,19 @@
+# View Object for an allocation history item
+
```
timpeat, line 2: Should we be using the presenter pattern now it's been added? (I'm ambivalent)

timpeat (approved):
Fantastic!

app/models/allocation_history.rb
hiboabd: Ah good spot! Removed

app/models/allocation_history_item.rb
willmcb: I think we should as it would be good to use one 'thing', rather than have separate concepts for essentially the same thing.

app/models/allocation_history.rb
hiboabd: Ah good spot! That seems best to me to keep `user_events` flexible. I'll make that change

app/views/manage_competencies/caseworker_competencies/_table_cells.html.erb
```diff
@@ -1,7 +1,12 @@
...
-  <% if Allocating.user_competencies(caseworker.id).blank? %>
-      <%= t('.no_competencies') %>
-  <% else %>
-    <%= Allocating.user_competencies(caseworker.id).map(&:humanize).join(', ') %>
-  <% end %>
+  <a class="govuk-link govuk-link--no-visited-state" href="<%=edit_manage_competencies_caseworker_competency_path(id: caseworker)%>">
+    <% if Allocating.user_competencies(caseworker.id).blank? %>
+        <%= t('.no_competencies') %>
+    <% else %>
+      <%= Allocating.user_competencies(caseworker.id).map(&:humanize).join(', ') %>
```
willmcb, line 6: If we do add a presenter then perhaps we could wrap this in a mothod which explaines it like `formatted_user_competencies` or something similar

app/views/manage_competencies/history/show.html.erb
```diff
@@ -0,0 +1,57 @@
...
+        <% @caseworker.competency_history.items.each do |event| %>
+          <tr class="govuk-table__row">
+            <td class="govuk-table__cell">
+              <%= l event.timestamp, format: :datetime %>
+            </td>
+            <td class="govuk-table__cell">
+              <% if event.data[:competencies].empty? %>
+                <%= t '.competencies_set', competencies_text: 'no competencies' %>
+              <% else %>
+                <%= t '.competencies_set', competencies_text: event.data[:competencies].map(&:humanize).join(', ') %>
```
willmcb, line 39: Probably could do similar here though not a big deal either way tbh

config/locales/en/manage_competencies.yml
```diff
@@ -10,11 +10,21 @@ en:
         heading: Manage competencies
       table_cells:
         no_competencies: No competencies
+        history: History
+    history:
+      show:
+        page_title: Competencies history
```
willmcb, line 16: This might be clearer as User's Competency History as thats what we are referring to in the heading.

app/views/manage_competencies/history/show.html.erb
hiboabd: Probably best to be consistent with the pattern we've just established. I need to get into the habit of using presenters 😄

PR #459, opened 2023-11-16
[CRIMAPP-93] Get next application by caseworker work stream

hiboabd:
## Description of change
Get next application assigns a caseworker an extradition case if they are assigned to an extradition work stream. Caseworkers that are not assigned to the extradition work stream will be assigned the next unassigned application regardless of work stream.

## Link to relevant ticket
[CRIMAPP-93](https://dsdmoj.atlassian.net/browse/CRIMAPP-93)

## Notes for reviewer
Note this only handles a caseworker with one competency (extradition)

### After changes:

**Caseworker competencies** 

Screenshot 2023-11-16 at 16 29 15

**Caseworker with no competencies set is assigned the next unassigned application** 

Screenshot 2023-11-16 at 16 28 53
Screenshot 2023-11-16 at 16 28 59

**Caseworker with extradition work stream set is assigned the next extradition application** 

Screenshot 2023-11-16 at 16 28 38
Screenshot 2023-11-16 at 16 28 32

Requires one supervisor user and two caseworker users (can also use one caseworker user if you want).
Also requires pulling down the datastore changes and spinning up review, apply and the datastore.

In apply, you'll need to submit an application with a first court hearing at Westminster Magistrates Court. You can also submit an application which did not have/is not having a first court hearing at Westminster Magistrates Court. 

Log in as a supervisor user in review and go to the manage competencies page, where you will set the competency of one of the caseworkers as extradition and the other caseworker can remain unset. 

Then log in as the caseworker assigned to the extradition work stream and go to assigned applications and click Get Next. You should be assigned to the application you just submitted with a first court hearing at Westminster Magistrates Court. (Or you may see an application you submitted earlier if you had an older unassigned application with a first court hearing at Westminster Magistrates Court.)

Then log in as the caseworker not assigned to the extradition work stream and go to assigned applications and click Get Next. You should not be assigned to the application you just submitted with a first court hearing at Westminster Magistrates Court. You should be assigned to the unassigned application that is next up in the queue. 

Play around with setting and unsetting the extradition competency and see if you can spot any weird behaviour

[CRIMAPP-93]: https://dsdmoj.atlassian.net/browse/CRIMAPP-93?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/get_next.rb
```diff
@@ -1,6 +1,6 @@
 class GetNext
-  def initialize
-    filter = ApplicationSearchFilter.new(assigned_status: 'unassigned')
+  def initialize(work_streams)
```
timpeat, line 2: Should we be using a hash rather than a positional argument here?

timpeat (commented):
One small change on the arguments for GetNext.

app/models/get_next.rb
```diff
@@ -9,4 +9,10 @@ def initialize
   def call
     @search.results.first&.id
   end
+
+  def self.call(work_streams:)
+    work_streams = [] if work_streams.nil?
```
mseedat-moj, line 14: I think you whittle down to `work_streams ||= []`

mseedat-moj (changes requested):

mseedat-moj (approved):
👊

PR #481, opened 2023-12-07
Crimapp 254 workstreams url changes

timpeat:
## Description of change

Allows WorkStream values to have different url params.

## Link to relevant ticket
[CRIMAPP-253](https://dsdmoj.atlassian.net/browse/CRIMAPP-253)

## Notes for reviewer
Introduces a WorkStream value object to allow urls to be independent of the work stream names stored in the database and events.

The names of the work streams have changed several times and it seems likely they might change again in the future. The introduction of the WorkStream value object allows presentation of WorkStream names in both text and url to be independent of stored values.

Builds on https://github.com/ministryofjustice/laa-review-criminal-legal-aid/pull/480 (which should be merged before these changes).

## How to manually test the feature

[CRIMAPP-253]: https://dsdmoj.atlassian.net/browse/CRIMAPP-253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #482, opened 2023-12-08
Re enable work stream ff on staging

hiboabd:
## Description of change

Re-enables work stream feature flag that was [previously disabled](https://github.com/ministryofjustice/laa-review-criminal-legal-aid/pull/481) on staging for regression testing.


willmcb (approved):
👍

PR #487, opened 2023-12-18
Handle sorting by integer when nil

timpeat:
Handle sorting by integer when nil

## Link to relevant ticket
https://ministryofjustice.sentry.io/issues/4729005239/?environment=staging&query=&referrer=issue-stream&statsPeriod=7d&stream_index=1

## Notes for reviewer

nil values are assigned -1 for sorting which means they are differentiated from 0%

Attempt to sort the Caseworker report on Review when one of the percentages has a nil value.


willmcb (approved):
👍

PR #494, opened 2023-12-19
[CRIMAPP-87] Allow Data Analyst casework read access

timpeat:
## Description of change
This pull request enables DataAnalysts to access the caseworker section of the service.

## Link to relevant ticket
https://dsdmoj.atlassian.net/jira/software/c/projects/CRIMAPP/boards/1375/backlog?selectedIssue=CRIMAPP-87

## Notes for reviewer

Previously, DataAnalysts didn't have access to the casework section. However, in practice, they require access to the search function and application details and history. Due to restrictions, DataAnalysts were included as Supervisors. Changes in competencies now permit users to access the caseworker section, but without specific caseworker competencies, they can't review applications.

This PR grants DataAnalysts access to the caseworker area, allowing them to view applications and conduct searches but preventing them from reviewing applications.

We do not have Data Analysts enrolled on production so this change can be deployed following regression testing.

Sign in as a data analyst. 
Confirm you are redirected to the application search. 
Confirm you cannot assign / review an application. 
Confirm you can access the reporting dashboard.
When on a report, allow the session to time out, confirm you are redirected to the report and not search when you re-authenticate.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -19,11 +19,8 @@ class User < ApplicationRecord
 
   before_create :set_invitation_expires_at
 
-  # Allow data_analyst users to be deleted because they cannot
-  # be assigned another 'lesser' role - see AddDataAnalystRoleEnum
-  # migration which destroys data_analysts if DB is rolled back
   before_destroy do
-    raise CannotDestroyIfActive if activated? && !data_analyst?
+    raise CannotDestroyIfActive if activated?
   end
 
   attr_readonly :email, :can_manage_others
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,12 +4,12 @@ class ApplicationController < ActionController::Base
   class ForbiddenError < StandardError; end
 
   def landing_page_for(user)
-    if user.service_user?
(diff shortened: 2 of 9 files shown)


app/models/user.rb
```diff
@@ -19,11 +19,8 @@ class CannotChangeRole < StandardError; end
 
   before_create :set_invitation_expires_at
 
-  # Allow data_analyst users to be deleted because they cannot
-  # be assigned another 'lesser' role - see AddDataAnalystRoleEnum
-  # migration which destroys data_analysts if DB is rolled back
   before_destroy do
-    raise CannotDestroyIfActive if activated? && !data_analyst?
+    raise CannotDestroyIfActive if activated?
```
timpeat, line 23: Data Analysts should not be deleted.

hiboabd (approved):
Works locally. Note my pr needs to be merged too to fix a bug found in QA.

Note discussed with Tim around removing the tabs data analysts don't need (your list, open/closed applications) but for the time being they are preventing from reviewing/viewing applications in the same way that caseworkers with no competencies are so this is not a priority

PR #496, opened 2023-12-20
[CRIMAPP-279] Ensure assigned applications are displayed for users with no competencies

hiboabd:
## Description of change
Discovered a bug in qa where when caseworkers with no competencies click to retrieve the next application, there is an error shown. This is because the search and sort filters were not being set when the page is being rendered with a flash message so the assigned applications table couldn't be populated. In future, assigned applications may be unassigned from caseworkers when their competencies are removed but this hasn't been agreed yet 

Also removes the application type column from following qa, as it was decided that this should be displayed once change in financial circumstances applications are coming through. Currently, every application is 'initial' so the column is not helpful 

## Link to relevant ticket
[CRIMAPP-279](https://dsdmoj.atlassian.net/browse/CRIMAPP-279)
[CRIMAPP-282](https://dsdmoj.atlassian.net/browse/CRIMAPP-282)

### Before changes:
Screenshot 2023-12-19 at 12 50 09

### After changes:
Screenshot 2023-12-20 at 12 47 51
Screenshot 2023-12-20 at 12 48 00

Sign in as a caseworker with a CAT 1 competency 
Assign an application to yourself
Sign out and sign in as a supervisor 
Change the competency of that caseworker to no competencies
Sign out and sign in as that caseworker
Go to the assigned applications page click review next application 
You should get a banner saying you cannot review applications without a work queue and you should still see your assigned application 

For the search results table change, you can log in as any user, make a search and verify that you can no longer see the application type column 

[CRIMAPP-279]: https://dsdmoj.atlassian.net/browse/CRIMAPP-279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CRIMAPP-282]: https://dsdmoj.atlassian.net/browse/CRIMAPP-282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/casework/assigned_applications_controller.rb
+++ b/app/controllers/casework/assigned_applications_controller.rb
@@ -11,10 +11,7 @@ module Casework
     def index
       return unless assignments_count.positive?
 
-      set_search(
-        default_filter: { assigned_status: current_user_id },
-        default_sorting: { sort_by: 'submitted_at', sort_direction: 'ascending' }
-      )
+      set_search_filters
     end
 
     def create
@@ -66,7 +63,16 @@ module Casework
       return unless current_user.work_streams.empty?
 
       set_flash(:no_work_streams_to_assign_from, success: false)
+      set_search_filters
+
       render :index
+    end
+
+    def set_search_filters
+      set_search(
+        default_filter: { assigned_status: current_user_id },
+        default_sorting: { sort_by: 'submitted_at', sort_direction: 'ascending' }
+      )
     end
   end
 end
--- a/app/controllers/casework/assigned_applications_controller.rb
+++ b/app/controllers/casework/assigned_applications_controller.rb
@@ -11,7 +11,10 @@ module Casework
     def index
       return unless assignments_count.positive?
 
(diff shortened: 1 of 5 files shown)


app/controllers/casework/assigned_applications_controller.rb
```diff
@@ -66,7 +63,16 @@ def require_a_user_work_stream
       return unless current_user.work_streams.empty?
 
       set_flash(:no_work_streams_to_assign_from, success: false)
-      render :index
+      set_search_filters
```
timpeat, line 66: @hiboabd I don't think you need to set search filters now this action has changed from using "render" to "redirect_to"

timpeat (approved):
Thanks for fixing @hiboabd. Sorry for introducing the regression!

PR #518, opened 2024-01-12
Revert importmap upgrade

timpeat:
## Description of change
Revert importmap upgrade

## Notes for reviewer
This was causing assets not to build correctly on live services. Temporary reversion to prevent this issue on live services.


hiboabd (approved):
Speedy fix, thank you!

PR #583, opened 2024-04-23
Conditionally show income sections for appeal no changes

arthurashman:
## Description of change
Conditionally show income sections for appeal no changes

Populate a sensible value for Application Mean Tested field
## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-754

### DIFF
--- a/app/views/casework/crime_applications/_client_details.html.erb
+++ b/app/views/casework/crime_applications/_client_details.html.erb
@@ -44,38 +44,40 @@
         <%= nino.presence || t(:not_provided, scope: 'values') %>
       </dd>
     </div>
-    <div class="govuk-summary-list__row">
-      <dt class="govuk-summary-list__key">
-        <%= label_text(:telephone_number) %>
-      </dt>
-      <dd class="govuk-summary-list__value">
-        <%= applicant_tel.presence || t(:not_provided, scope: 'values') %>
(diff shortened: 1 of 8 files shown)


app/views/casework/crime_applications/_client_details.html.erb
```diff
@@ -44,38 +44,40 @@
...
-      </dt>
-      <dd class="govuk-summary-list__value">
-        <% if applicant.correspondence_address_type == 'other_address' %>
-          <%= render 'address', address: applicant.correspondence_address %>
-        <% else %>
-          <%= t(applicant.correspondence_address_type, scope: 'values') %>
-        <% end %>
-      </dd>
-    </div>
+    <% unless crime_application.case_details.appeal_reference_number %>
```
hiboabd, line 47: What if a usn is provided rather than a reference number?

arthurashman: appeal_reference_number field is the type, so either "appeal_usn" or "appeal_maat_id" then there are fields with those names with the actual ids in.

hiboabd: Ahhh yes 😄 thanks

hiboabd (approved):

PR #613, opened 2024-06-07
Upgrade rails and devise

timpeat:
## Description of change

Upgrade rails 7.3.4 and upgrade devise 4.9.4

## Notes for reviewer

I have tested and confirm that the authentication via azure_ad still works after this update.


hiboabd (approved):
👍

PR #696, opened 2024-09-03
Increase sample period for 5xx and 4xx alerts

timpeat:
## Description of change

Increase the sampling period for 4xx and 5xx prometheus alerts.

## Notes for reviewer

Updates the alerts to triggers if the total rate exceeds 5 errors per minute, but averaged over a 5-minute period to hopefully reduce number of unnecessary alerts.


arthurashman (approved):
Ta 👍

PR #720, opened 2024-10-09
LAA-REVIEW-CRIMINAL-LEGAL-AID-1M

timpeat:
## Description of change

Handle missing partner address

## Link to relevant ticket

https://dsdmoj.atlassian.net/browse/CRIMAPP-1386

## Notes for reviewer

Urgent bug fix.

### DIFF
--- a/spec/system/casework/viewing_an_application/application_details/partner_details_spec.rb
+++ b/spec/system/casework/viewing_an_application/application_details/partner_details_spec.rb
@@ -92,17 +92,26 @@ RSpec.describe 'When viewing partner details' do
       end
 
       context 'when home address is present' do
-        subject(:home_address) { page.find('dt', text: 'Home address').find('+dd') }
-
-        let(:application_data) do
-          super().deep_merge('client_details' =>
-                               { 'applicant' => { 'residence_type' => 'none',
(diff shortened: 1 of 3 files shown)


hiboabd (approved):
Thank you!

spec/system/casework/viewing_an_application/application_details/partner_details_spec.rb
```diff
@@ -92,17 +92,26 @@
...
-                               { 'applicant' => { 'residence_type' => 'none',
-                                                  'relationship_to_owner_of_usual_home_address' => nil } })
+                               { 'partner' => { 'has_same_address_as_client' => 'no',
+                                                                  'home_address' => nil } })
         end
 
         it "displays the partner's address details" do
           expect(page).to have_content('Lives at same address as client? No')
-          expect(home_address).to have_content('53 Road Street Another nice city W1 2AA United Kingdom')
+          expect(home_address).to have_content ''
```
hiboabd, line 114: This should be not to have content as the row wouldn't appear

timpeat: This scenario exists because the home address is missing, so I'd rather leave it in. (Unlike the other scenarios where it is not required.)

PR #724, opened 2024-10-14
Update search page default review status to all applications

hiboabd:
## Description of change
Changes search default from open to all applications. This was attempted in PR #722 but there were areas of the app that returned all applications when they shouldn't have been. So this PR also explicitly sets the review status search filter to `open` for assigned applications and get next functions 

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1385

### After changes:
Screenshot 2024-10-10 at 08 46 13

### DIFF
--- a/app/lib/types.rb
+++ b/app/lib/types.rb
@@ -18,6 +18,7 @@ module Types # rubocop:disable Metrics/ModuleLength
   # Map of review status groups to LaaCrimeSchemas::Types:REVIEW_APPLICATION_STATUSES
   #
   REVIEW_STATUS_GROUPS = {
+    'all' => REVIEW_APPLICATION_STATUSES,
     'open' => [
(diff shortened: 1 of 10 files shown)


app/lib/types.rb
```diff
@@ -30,10 +31,9 @@ module Types # rubocop:disable Metrics/ModuleLength
     'sent_back' => [
       Types::ReviewApplicationStatus['returned_to_provider']
     ],
-    'all' => REVIEW_APPLICATION_STATUSES
   }.freeze
 
-  ReviewStatusGroup = String.default('open'.freeze).enum(
+  ReviewStatusGroup = String.default('all'.freeze).enum(
```
timpeat, line 36: With hindsight, I think defaulting here in the Types should be avoided.

timpeat (approved):

PR #725, opened 2024-10-15
Fix unhandled errors in system specs

timpeat:
## Description of change

Raise unhandled errors in system specs

## Notes for reviewer

The `show_exceptions` setting was incorrectly configured to `:all` in the test environment, causing unhandled exceptions (e.g., WebMock errors) in system specs to be hidden.

Changing the configuration to `:rescuable` ensures that handled error pages are shown, while unhandled errors are properly raised.

### DIFF
--- a/spec/system/casework/viewing_an_application/application_details/supporting_evidence_spec.rb
+++ b/spec/system/casework/viewing_an_application/application_details/supporting_evidence_spec.rb
@@ -2,6 +2,7 @@ require 'rails_helper'
 
 RSpec.describe 'Viewing supporting evidence' do
   include_context 'with stubbed application'
+  include_context 'when downloading a document'
 
   before do
     visit crime_application_path(application_id)
--- a/spec/system/errors_spec.rb
+++ b/spec/system/errors_spec.rb
@@ -2,27 +2,6 @@ require 'rails_helper'
 
 RSpec.describe 'Error pages' do
   context 'when authenticated' do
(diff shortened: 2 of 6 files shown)


spec/system/casework/viewing_an_application/application_details/supporting_evidence_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 RSpec.describe 'Viewing supporting evidence' do
   include_context 'with stubbed application'
+  include_context 'when downloading a document'
```
hiboabd, line 5: Why has this changed? is this because there was an unhandled exception?

spec/system/errors_spec.rb
```diff
@@ -2,27 +2,6 @@
 
 RSpec.describe 'Error pages' do
   context 'when authenticated' do
-    context 'when an unhandled error occurs' do
```
hiboabd, line 5: Why has this spec been removed?

spec/system/casework/viewing_an_application/application_details/supporting_evidence_spec.rb
timpeat: Yes

spec/system/errors_spec.rb
timpeat: Testing this in the system spec is no longer possible with the show_exception configuration as it is.

hiboabd (approved):

PR #759, opened 2024-11-20
CRIMAPP-1419 support fail means and IoJ

timpeat:
## Description of change

- Support 'failmeioj'
- Update decision summary cards to latest content
- Make decision results consistent

## Link to relevant ticket

Fixes bug found in: https://dsdmoj.atlassian.net/browse/CRIMAPP-1473
Updates content as per: https://dsdmoj.atlassian.net/browse/CRIMAPP-1419


app/components/decision_component.rb
```diff
@@ -61,15 +61,15 @@ def update_path
   end
 
   def ioj_result
-    t(interests_of_justice.result, scope: [:values, :decision_result])
```
timpeat, line 64: The same values (e.g. fail) are shown as different values depending on test. (E.g. Refused v Failed)

app/components/decision_result_component.rb
```diff
@@ -22,7 +22,7 @@ def call
 
   def colour
     case result
-    when /fail/
+    when /fail|inel/
```
timpeat, line 25: I'm still not sure that inel will be sent by the MAAT API, but it could theoretically, so supporting it now.

app/lib/maat/decision.rb
```diff
@@ -28,11 +28,22 @@ def funding_decision(maat_value)
         maat_value.downcase
       end
 
+      def corrected_ioj_result(response)
```
timpeat, line 31: The value updated value of the IoJ decision is not currently being sent. Here we reverse engineer the result.

EdwinKruglov (approved):

PR #771, opened 2024-12-10
[CRIMAPP-1508] Summary card content updates

EdwinKruglov:
## Description of change
Made content changes to the summary cards as per the latest design. These changes bring the summary content of the Application Details page in line with Crime Apply.

## Link to relevant ticket
[CRIMAPP-1508](https://dsdmoj.atlassian.net/browse/CRIMAPP-1508)
[Figma design](https://www.figma.com/design/thkvnDBbQbHlSqzt1TE3lk/Crime-Apply?node-id=19698-65906&t=Dl32HUvXFHf6vQOF-1)

[CRIMAPP-1508]: https://dsdmoj.atlassian.net/browse/CRIMAPP-1508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/locales/en/casework.yml
```diff
@@ -131,16 +130,17 @@ en:
     case_type: Case type
     caseworker: Caseworker
     client_details: Client details
+    client_contact_details: Client contact details
     client_in_armed_forces: Client in armed forces?
     client_other_charges: Other charges
     client_other_charges_explicit: "Other charges: client"
-    client_owns_property: Has land or property?
+    client_owns_property: '%{prefix} owns home, land or property?'
```
hiboabd, line 137: Not a big deal but i think subject would be better than prefix

spec/system/casework/viewing_an_application/application_details/partner_details_spec.rb
```diff
@@ -124,8 +123,7 @@
 
         it "displays the partner's address details" do
           expect(summary_card('Partner details')).to have_rows(
-            'Lives at same address as client?', 'No',
-            'Home address', ''
+            'Lives at same address as client?', 'No'
```
hiboabd, line 126: If they don't live at the same address as the client shouldn't the home address show? or is an address not stubbed?

app/views/casework/crime_applications/_first_court_hearing.html.erb
```diff
@@ -0,0 +1,14 @@
+<% return if case_details.first_court_hearing.nil? %>
```
hiboabd, line 1: Should this file and other newer partials be in the `sections/` folder

app/helpers/app_text_helper.rb
```diff
@@ -15,6 +15,17 @@ def label_text(key, options = {})
     named_text(:labels, key, options)
   end
 
+  def label_text_with_subject(key, subject, options = {}, subject_appended: true)
```
hiboabd, line 18: I think the label text helper does handle subject. Why is this necessary?

spec/system/casework/viewing_an_application/application_details/partner_details_spec.rb
EdwinKruglov: The home address is nil in this context, so the `Home address` row will not be visible. There's a test just above this one that tests the scenario you describe.

app/helpers/app_text_helper.rb
EdwinKruglov: It does, but I needed a bit more flexibility to cover scenarios that deal with subjects. For many of the summary boxes we want to display the subject, but only if a partner exists. So I decided to be explicit about it with this method. I suppose in most cases I could've added two versions of each title to the locales, e.g. "Payments" and "Payments: %{subject}", but I thought just having one and appending the subject there might be nicer?

There's also one case where the wording is different (plural) if we're dealing with a client and a partner: `_other_income_details.html.erb:6` and `casework.yml:261`, so I wanted to account for that in the same method as well.

hiboabd (approved):

app/helpers/app_text_helper.rb
hiboabd: I see, thanks!

PR #774, opened 2024-12-11
[CRIMAPP-1510] Sanitise upload file names

EdwinKruglov:
## Description of change
Update `response_content_disposition` for uploaded files to handle file names that contain characters outside the ISO-8859-1 character set.

## Link to relevant ticket
[CRIMAPP-1510](https://dsdmoj.atlassian.net/browse/CRIMAPP-1510)

Upload a file whose name contains UTF-8 characters on Apply and attempt to download them on Review.
Some examples that would normally fail:
- `Document 1 – Main Account 1-15 November.pdf`
- `Payslip - Sept ‘24.pdf`

[CRIMAPP-1510]: https://dsdmoj.atlassian.net/browse/CRIMAPP-1510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/datastore/documents/download.rb
+++ b/app/services/datastore/documents/download.rb
@@ -30,7 +30,9 @@ module Datastore
 
       def response_content_disposition
         # To force download of file rather than opening in another window
-        %(attachment; filename="#{@document.filename}")
+        filename_safe = @document.filename.gsub(/[^a-zA-Z0-9._-]/, '_')
+        filename_escaped = ERB::Util.url_encode(@document.filename)
+        %(attachment; filename=#{filename_safe}; filename*= UTF-8''#{filename_escaped};)
       end
     end
   end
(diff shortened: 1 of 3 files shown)


timpeat (approved):
After discussing this PR with @EdwinKruglov, we agreed that sanitising the file name in the downloader's response, rather than during file upload, is the better approach for now. The key reasons are:  

1. It works for all existing files without requiring a migration.  
2. Security risks from file names are already mitigated within our internal systems.  
3. Storing the uploader's original file name format ensures we can address issues where sanitisation might cause collisions during download.  
4. Any issues with the sanitisation process can be addressed without requiring a data migration. For example, if the sanitisation logic strips critical information, we can adjust the filename regex used during download without having to migrate data or update the TTL on existing files.

PR #789, opened 2025-01-13
Crimapp 1534 no empty cells

timpeat:
## Description of change
Add visually hidden text to empty table cells.

## Link to relevant ticket

[CRIMAPP-1534](https://dsdmoj.atlassian.net/browse/CRIMAPP-1534)

## Notes for reviewer

As per accessibility recommendation, add visually hidden text to in format "No data exists for %{column_name}" to empty table cells.


app/views/casework/base/_primary_navigation.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= govuk_service_navigation(navigation_id: 'example-5', current_path: request.path, classes: 'app-service-navigation') do |sn| %>
+<%= govuk_service_navigation(current_path: request.path, classes: 'app-service-navigation') do |sn| %>
```
timpeat, line 1: ☝️ removes unrelated copy paste error spotted as part of this work.

app/views/casework/application_searches/_search_results_table_body.html.erb
EdwinKruglov, line 1: Looks like case type could be blank as well. Should we be handling empty cells in that column?

timpeat: Good call. Thanks @EdwinKruglov

EdwinKruglov (approved):

PR #793, opened 2025-01-16
[CRIMAPP-1542] Add margin to assignment buttons

hiboabd:
## Description of change

Updated margin for assignment buttons due to feedback from accessibility review: 
`There needs to be enough space between buttons and other text, so that users who do not have good dexterity can select interactable elements easily.`

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1542

## Notes for reviewer
A number of buttons in the service could probably do with reviewing and it's not clear how much spacing is required. This pr is only considering what has been flagged in the ticket and is essentially a guesstimate as to how much spacing is required

### Before changes:
**At 300% zoom**
Screenshot 2025-01-16 at 15 44 29

### After changes:
**At 300% zoom**
Screenshot 2025-01-16 at 15 42 09


EdwinKruglov (approved):
Good point about the lack of guidance around spacing. When you zoom in at 250% the bottom margin of the text above loses 10px, so adding 10px to the button instead and keeping that consistent sounds like a good compromise to me!

PR #810, opened 2025-02-10
[CRIMAPP-1600] Update non means submit decision form

hiboabd:
## Description of change
Updates submit decision form for non means applications following feedback from user research 

Changes:
- hide ‘What do you want to do next?’ and the radio buttons block
- replace ‘Save and continue’ button with a button saying ‘Send decision to provider’ 
- hide H2 ‘You have added 1 case’

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1600

### Before changes:
Screenshot 2025-02-10 at 09 41 49

### After changes:
Screenshot 2025-02-10 at 09 41 25

Submit a non means application and then start adding the funding decision on review 
After adding any decision comments you should see a screen with a summary card of the decision and a button to submit the decision to the provider 
Verify the changes listed above have been made and then confirm that you can submit a non means decision

### DIFF
--- a/app/models/crime_application.rb
+++ b/app/models/crime_application.rb
@@ -3,6 +3,7 @@ require 'laa_crime_schemas'
 class CrimeApplication < LaaCrimeSchemas::Structs::CrimeApplication # rubocop:disable Metrics/ClassLength
   include Assignable
   include Reviewable
+  include TypeOfApplication
 
   PRE_CIFC_MAAT_ID = 'pre_cifc_maat_id'.freeze
   PRE_CIFC_USN = 'pre_cifc_usn'.freeze
@@ -188,19 +189,6 @@ class CrimeApplication < LaaCrimeSchemas::Structs::CrimeApplication # rubocop:di
 
   def date_stamp_context
     @date_stamp_context ||= DateStampContextPresenter.present(DateStampContext.new(self))
-  end
-
-  def pse?
-    application_type == Types::ApplicationType['post_submission_evidence']
-  end
-
-  def cifc?
-    application_type == Types::ApplicationType['change_in_financial_circumstances']
-  end
-
-  def appeal_no_changes?
-    case_details&.case_type == Types::CaseType['appeal_to_crown_court'] &&
-      case_details&.appeal_reference_number.present?
   end
 
   def last_jsa_appointment_date?(person)
--- a/app/views/casework/send_decisions/new.html.erb
+++ b/app/views/casework/send_decisions/new.html.erb
@@ -29,9 +29,7 @@
       <%= form_with model: @form_object, url: crime_application_send_decisions_path do |f| %>
 
         <% if @crime_application.work_stream == 'non_means_tested' %>
-          <%= f.govuk_radio_buttons_fieldset :next_step, legend: { hidden: true }, hidden: true do %>
-            <%= f.govuk_radio_button :next_step, :send_to_provider, checked: true, hidden: true %>
-          <% end %>
+          <%= f.hidden_field :next_step, value: :send_to_provider %>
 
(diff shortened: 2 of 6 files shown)


spec/system/casework/deciding/adding_another_maat_decision_spec.rb
```diff
@@ -63,6 +63,13 @@
     expect(current_path).to eq(crime_application_send_decisions_path(application_id))
   end
 
+  context 'when "what do you want to do next?" not answered' do
```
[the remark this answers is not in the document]
hiboabd: Good point - would be good to catch any regressions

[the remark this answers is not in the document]
hiboabd: Hm non means applications would not go through this route, as you don't add a decision via a MAAT ID so I wouldn't be able to add a context block in this spec. I could add it to the `submitting_a_non_means_decision_spec.rb`  spec but the confirmation spec already covers this and checks a success banner is shown so overall don't think it's necessary :)

app/views/casework/send_decisions/new.html.erb
```diff
@@ -26,16 +28,27 @@
     <% if @decisions.all? &:complete? %>
       <%= form_with model: @form_object, url: crime_application_send_decisions_path do |f| %>
 
-        <%= f.govuk_radio_buttons_fieldset :next_step, legend: { tag: 'h2', size: 'm' } do %>
-
-          <%= f.govuk_radio_button :next_step, :send_to_provider, link_errors: true %>
-          <%= f.govuk_radio_button :next_step, :add_another %>
+        <% if @crime_application.work_stream == 'non_means_tested' %>
```
timpeat, line 31: I'm wondering if its time for a "TypeOfApplication" concern that include as we're defining this in a few places:

https://github.com/ministryofjustice/laa-review-criminal-legal-aid/blob/6e8daffb77cd4305f5db4a2f9399bba24563d06a/app/aggregates/reviewing/available_reviewer_actions.rb#L65-L75

https://github.com/ministryofjustice/laa-review-criminal-legal-aid/blob/6e8daffb77cd4305f5db4a2f9399bba24563d06a/app/models/crime_application.rb#L193-L199

app/views/casework/send_decisions/new.html.erb
```diff
@@ -26,16 +28,27 @@
...
+          <%= f.govuk_radio_buttons_fieldset :next_step, legend: { hidden: true }, hidden: true do %>
```
timpeat, line 32: Would you mind changing this to use a hidden_field input rather than hiding these with css? (e.g. https://api.rubyonrails.org/v7.0/classes/ActionView/Helpers/FormHelper.html#method-i-hidden_field )

app/models/crime_application.rb
```diff
@@ -190,19 +191,6 @@ def date_stamp_context
     @date_stamp_context ||= DateStampContextPresenter.present(DateStampContext.new(self))
   end
 
-  def pse?
```
timpeat, line 193: ❤️

app/views/casework/send_decisions/new.html.erb
```diff
@@ -11,9 +11,11 @@
       <%= @crime_application.applicant.full_name %>
     </h1>
 
-    <h2 class="govuk-heading-m">
-      <%= t '.title', count: @decisions.size %>
-    </h2>
+    <% unless @crime_application.work_stream == 'non_means_tested' %>
```
timpeat, line 14: You can use your new concern ☝️ `@crime_application.non_means?`

hiboabd: I forgot 😭 thank you

timpeat (approved):
Thank you!

PR #815, opened 2025-02-12
[CRIMAPP-1641] Show assets based on properties array

EdwinKruglov:
## Description of change
This ensures that the client's assets are shown based on the properties they've declared. This addresses an issue where some applications will have both `has_no_properties: "yes"` and declared properties due to a bug on Apply.

## Link to relevant ticket
[CRIMAPP-1641](https://dsdmoj.atlassian.net/browse/CRIMAPP-1641)

[CRIMAPP-1641]: https://dsdmoj.atlassian.net/browse/CRIMAPP-1641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/casework/crime_applications/sections/_properties.html.erb
+++ b/app/views/casework/crime_applications/sections/_properties.html.erb
@@ -2,7 +2,7 @@
   <h2 class="govuk-heading-l"><%= label_text(:assets) %></h2>
 <% end  %>
 
-<% if capital_details.has_no_properties == 'yes'  %>
+<% if capital_details.properties.empty?  %>
   <%= govuk_summary_card(title: label_text(:properties_title)) do %>
     <%= govuk_summary_list(actions: false) do |list|
           list.with_row do |row|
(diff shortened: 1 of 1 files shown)


app/views/casework/crime_applications/sections/_properties.html.erb
```diff
@@ -2,7 +2,7 @@
   <h2 class="govuk-heading-l"><%= label_text(:assets) %></h2>
 <% end  %>
 
-<% if capital_details.has_no_properties == 'yes'  %>
+<% if capital_details.properties.empty?  %>
```
hiboabd, line 5: I just realised when there is a partial capital assessment, this section will show when it should not 🥲. I think we may need to introduce a helper method that can determine if the full means assessment has been undertaken? Happy to discuss

timpeat: If properties are not relevant to the application at submission, they should not be included in the data that is sent to the datastore. Happy to help work out how to stop that form happening.

timpeat (approved):

PR #828, opened 2025-03-11
[CRIMAPP-1532] Screenreader announcement for table

lucas-shaw:

### DIFF
--- a/app/views/casework/application_searches/search.html.erb
+++ b/app/views/casework/application_searches/search.html.erb
@@ -20,7 +20,7 @@
 
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full app-table-container">
-      <table class="govuk-table app-table">
+      <table class="govuk-table app-table" aria-roledescription="." aria-label="Table, <%= t('.total', count: @search.total) %>">
         <%= render DataTable::HeadComponent.new(sorting: @search.sorting, filter: @search.filter) do |head| %>
           <% head.with_row do |row|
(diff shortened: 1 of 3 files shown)


app/views/casework/application_searches/search.html.erb
```diff
@@ -20,7 +20,7 @@
 
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full app-table-container">
-      <table class="govuk-table app-table">
+      <table class="govuk-table app-table" aria-roledescription="." aria-label="Table, <%= t('.total', count: @search.total) %>">
```
lucas-shaw, line 23: Added in `aria-roledescription="."` to overwrite the default of saying "table" at the end. As it would read out as "Entering table, x search result table" which doesnt sound right.

app/views/casework/application_searches/search.html.erb
```diff
@@ -20,7 +20,7 @@
 
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full app-table-container">
-      <table class="govuk-table app-table">
+      <table class="govuk-table app-table" aria-roledescription="." aria-label="<%= t('.table_label') %> <%= t('.total', count: @search.total) %>">
```
lucas-shaw, line 23: Added `aria-roledescription="."` because otherwise the announcement uses the default "table" at the end, which would read "Entering application search results table, n search results table..."

EdwinKruglov (approved):

PR #835, opened 2025-03-18
[CRIMAPP-1295] Caseworker reports show empty %

lucas-shaw:
## Description of change
This issue is prevalent on all 3 percentage columns if the the guard clause return true. Changing `nil` to `0`ensures a value will be displayed

## Link to relevant ticket
https://dsdmoj.atlassian.net/jira/software/c/projects/CRIMAPP/boards/1375?selectedIssue=CRIMAPP-1295

### After changes:
Screenshot 2025-03-18 at 09 25 18


timpeat (approved):
Nice

PR #858, opened 2025-04-25
[CRIMAPP-1679] Add Sidekiq

EdwinKruglov:
## Description of change
This change adds Sidekiq to process background jobs. This initial configuration is to be tested in staging in preparation for work to implement automatically generated monthly reports. As a first test, this change should make it so that `application_returned_email` jobs are delegated to the `mailers` queue by Sidekiq.

This relies on an ElastiCache (Redis) instance deployed on CloudPlatform.
I've added Prometheus metrics that we could use to track and set up alerts. We could also set up a Grafana dashboard, as NSCC have done [here](https://grafana.live.cloud-platform.service.justice.gov.uk/d/laa-submit-crime-forms-sidekiq-metrics/laa-non-standard-crime-claims-sidekiq-metrics?from=now-24h&to=now&timezone=browser&var-namespace=laa-assess-crime-forms-prod&var-job=SendEmailToProvider&refresh=1m), though Sidekiq comes with its own web UI we could use, but we'd need to think about how to manage access to it in production.

I tried to keep the Kubernetes setup basic and minimal. I saw some examples of pretty involved startup probe checks, but I'd like to see how the basic one I have works before making it more complex.

I've also pinned the `sidekiq` gem version to 7 as `sidekiq-scheduler`, which we may use to schedule monthly jobs, does not currently support version 8.

## Link to relevant ticket

[CRIMAPP-1679]: https://dsdmoj.atlassian.net/browse/CRIMAPP-1679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -2,6 +2,13 @@ Rails.application.routes.draw do
   Rails.application.routes.draw { mount RailsEventStore::Browser => "/res" if Rails.env.development? }
   mount DatastoreApi::HealthEngine::Engine => '/datastore'
 
+  # TODO: should this be accessible in prod? By which users?
+  unless Rails.env.production?
+    require "sidekiq/web"
+    require 'sidekiq-scheduler/web'
+    mount Sidekiq::Web => "/sidekiq"
+  end
+
   get :health, to: 'healthcheck#show'
   get :ping,   to: 'healthcheck#ping'
 
--- a/.env.development
+++ b/.env.development
@@ -34,3 +34,5 @@ PORT=3001
 # If set, enables prometheus middleware and server
 # ENABLE_PROMETHEUS_EXPORTER=true
 # PROMETHEUS_EXPORTER_VERBOSE=false
+
+REDIS_URL=redis://localhost:6379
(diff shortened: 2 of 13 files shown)


config/routes.rb
```diff
@@ -2,6 +2,13 @@
   Rails.application.routes.draw { mount RailsEventStore::Browser => "/res" if Rails.env.development? }
   mount DatastoreApi::HealthEngine::Engine => '/datastore'
 
+  # TODO: should this be accessible in prod? By which users?
```
hiboabd, line 5: I think access should be internal, just wondering what it would display that other prod users would need? Also, is there an example on what it would look like/display?

EdwinKruglov: It would show information about how our jobs are doing, which ones are failing, etc. Lots of useful information in a nice UI. It would also allow you to easily kill and re-run jobs, so it would only be useful to and should be restricted to devs, realistically. A lot of that information would surface in the logs anyway and we can use Prometheus metrics to monitor it as well, so it's not essential but more of a convenience. I've seen other teams use basic username and password authentication.

image

hiboabd: Nice! probably useful for devs only to begin with and then we can give access to others if we feel it's necessary later

hiboabd (approved):

PR #876, opened 2025-05-14
[CRIMAPP-1679] Limit access to Sidekiq Web

EdwinKruglov:
## Description of change
This change limits access to the Sidekiq Web UI to admin/user managers in staging and locally. There will be no way to access it in hosted production.

## Link to relevant ticket

[CRIMAPP-1679]: https://dsdmoj.atlassian.net/browse/CRIMAPP-1679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -3,9 +3,11 @@ Rails.application.routes.draw do
   mount DatastoreApi::HealthEngine::Engine => '/datastore'
 
   unless HostEnv.production?
-    require 'sidekiq/web'
-    require 'sidekiq-scheduler/web'
-    mount Sidekiq::Web => "/sidekiq"
+    authenticate :user, lambda { |u| u.admin? } do
+      require 'sidekiq/web'
+      require 'sidekiq-scheduler/web'
+      mount Sidekiq::Web => "/sidekiq"
+    end
   end
 
   get :health, to: 'healthcheck#show'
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -3,9 +3,11 @@
   mount DatastoreApi::HealthEngine::Engine => '/datastore'
 
   unless HostEnv.production?
-    require 'sidekiq/web'
-    require 'sidekiq-scheduler/web'
-    mount Sidekiq::Web => "/sidekiq"
+    authenticate :user, lambda { |u| u.admin? } do
```
hiboabd, line 6: User manager and admin have the same definition, why not use user manager? or do we plan to differentiate in the future?

spec/models/concerns/user_role_spec.rb
```diff
@@ -269,4 +269,13 @@
       end
     end
   end
+
+  describe '#admin?' do
+    context 'with can_manager_others attribute set to true' do
```
hiboabd, line 274: Small typo: `can_manage_others`

hiboabd (approved):
Looks good, just couple of non-blocking comments

config/routes.rb
EdwinKruglov: I think admin makes more semantic sense here, than user_manager, and it might be useful to differentiate if we wanted to change it. Especially because in this context we would only want the devs to access Sidekiq, I think the intention is clearer if it's admin.

PR #877, opened 2025-05-14
[CRIMAPP-1679] Show links for generated reports

EdwinKruglov:
## Description of change
This change removes the temporary download section from /reporting and instead shows download links for generated reports under each report's table, if any are available, following the existing pattern with live reports.

## Link to relevant ticket

## Screenshots of changes
image


hiboabd (approved):
👏

PR #922, opened 2025-09-10
[CRIMAPP-1930] configure service hours in staging

timpeat:
## Description of change

Return 503 when provider service is accessed outside of office hours in staging

Staging is available 7am to 10pm seven days a week.

## Link to relevant ticket

CRIMAPP-1929

## Notes for reviewer

I've used @pm (pattern match) instead of `@within' partly because it allows easier testing of access control logic--particularly when simulating access during office hours without changing the rule structure--and partly because matching time to integer was sketchy.

I haven't yet verified whether values like 09, 08, etc., match correctly due to potential issues with leading zeros in numeric comparisons. If needed, I can adjust those to 9, 8, etc.


config/kubernetes/staging/ingress.yml
```diff
@@ -71,6 +71,8 @@ metadata:
       SecRuleEngine On
       SecRequestBodyLimit 1048576
       SecRequestBodyNoFilesLimit 1048576
+      SecRule TIME_HOUR "!@pm 06 07 08 09 10 11 12 13 14 15 16 17 18 19 20" \
```
hiboabd, line 74: Should this go to 21?

timpeat: Last request is 21:59:59, server is UTC so 20:59:59

PR #960, opened 2025-12-16
[CRIMAPP-2029] Ghost applications

EdwinKruglov:
## Description of change
- prohibit users that are not assigned to a given application from performing Reviewing actions: sending the application back, marking it as ready, or adding a funding decision.

## Link to relevant ticket
[CRIMAPP-2029](https://dsdmoj.atlassian.net/browse/CRIMAPP-2029)

## Notes for reviewer

In the specs I've introduced the helper `with_assignment` and the shared context `with stubbed assignment`. Both do the same thing, but in some of our specs it's enough to use the shared context and apply the stub to the entire context/example, whereas in others we may want it to be more granular and only apply it to a specific command call, which is where I've used the helper method.

I've chosen not to rely on stubbing in the system specs to make those scenarios more faithful to how the system would behave in the real world.

[CRIMAPP-2029]: https://dsdmoj.atlassian.net/browse/CRIMAPP-2029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/locales/en/casework.yml
+++ b/config/locales/en/casework.yml
@@ -64,8 +64,9 @@ en:
         - "%{continue_adding_decision_link}"
       continue_adding_decision: Continue adding funding decision
       unexpected_assignee:
-        - Action could not be completed
-        - This application is assigned to another user.
+        - You cannot review this application
+        - It has been reassigned to another team member.
+        - Contact your supervisor if you need to work on this application.
 
   primary_navigation:
     your_list: "Your list (%{count})"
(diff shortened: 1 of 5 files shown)


hiboabd (approved):
LGTM!

timpeat (approved):
Thank you :)

PR #970, opened 2026-01-14
Auditor role

lucas-shaw:
## Link to relevant ticket
https://dsdmoj.atlassian.net/jira/software/c/projects/CRIMAPP/boards/1375?assignee=712020%3A3d805bbe-8fa6-48b0-b8bd-f2d3612405d6&selectedIssue=CRIMAPP-2042

### After changes:
Screenshot 2026-01-19 at 13 48 10

For local testing, register then sign in with the new `auditor` role and confirm the roles access
`User.create(email: 'test.auditor@test.com', first_name: 'Test', last_name: 'Auditor', role: 'auditor')`

### DIFF
--- a/app/models/concerns/user_role.rb
+++ b/app/models/concerns/user_role.rb
@@ -42,10 +42,6 @@ module UserRole
     can_manage_others?
   end
 
-  def readonly_user?
-    role == AUDITOR && non_manager?
-  end
-
   def available_roles
     Types::USER_ROLES - [role]
   end


app/models/concerns/user_competence.rb
```diff
@@ -9,13 +9,11 @@ def work_streams
 
   # Types of applications a user can process
   def application_types_competencies
-    return [] if data_analyst?
```
lucas-shaw: Covered below in `def competencies`

app/models/concerns/user_role.rb
```diff
@@ -6,69 +6,72 @@ module UserRole
...
-
-    false
+    reporting_user? || user_manager_with_service_access?
   end
 
   def can_download_reports?
     role == DATA_ANALYST
   end
 
-  # TODO: Any reason to not allow supervisor to be 'downgraded' to caseworker?
```
lucas-shaw: was added 3 years ago

app/controllers/application_controller.rb
```diff
@@ -4,11 +4,11 @@ class ApplicationController < ActionController::Base
   class ForbiddenError < StandardError; end
 
   def landing_page_for(user)
-    if user.data_analyst?
+    if user.data_analyst? || user.auditor?
       new_application_searches_path
     elsif user.service_user?
       assigned_applications_path
-    elsif user.user_manager?
+    elsif user.can_manage_others?
```
timpeat: I think it’s best to keep the name as user_manager, since this method is about the type of user rather than what they can do (e.g., data_analyst, auditor, caseworker, supervisor). The user_manager role also comes with important restrictions--for example, a user manager can never switch to another user type or access the service in production. The data model makes this confusing for historical reasons.

app/models/concerns/user_role.rb
```diff
@@ -6,69 +6,72 @@ module UserRole
...
+  AUDITOR = Types::AUDITOR_ROLE
+  REPORTING_ROLES = [SUPERVISOR, DATA_ANALYST].freeze
+  SERVICE_USERS = [CASEWORKER, SUPERVISOR, DATA_ANALYST, AUDITOR].freeze
 
   included do
     # NOTE: mapping to PostgreSQL enum type via dry-types definition
     enum :role, Types::UserRole.mapping
   end
 
-  def can_read_application?
```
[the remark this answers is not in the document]
timpeat: Agree. I also think "can_write_application?" should go. Competencies and not roles are used to make these determinations. "can_write.." is used in the header. That use should be replaced with 'supervisor?'

app/models/concerns/user_role.rb
```diff
@@ -6,69 +6,72 @@ module UserRole
...
   end
 
-  # TODO: Any reason to not allow supervisor to be 'downgraded' to caseworker?
   def can_change_role?
-    return false unless activated?
-    return false if deactivated?
-    return false if dormant?
-
-    true
+    activated? && !deactivated? && !dormant?
```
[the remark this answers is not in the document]
timpeat: Personally I found it more readable before. Happy to go with majority though.

timpeat (approved):

PR #1176, opened 2026-07-10
CRIMAPP-2182: Reinstate sticky report headers

amazimbe:
## Description of change
- correct the sortable table HTML structure by moving the caption to the table level
- reduce report column widths so tables no longer require horizontal scrolling
- reinstate sticky headers now that the report scrolls with the page

## Link to relevant ticket
CRIMAPP-2182

Screenshot 2026-07-10 at 10 37 19

### Manual testing
Visit the reporting pages at [https://staging.review-criminal-legal-aid.service.justice.gov.uk/reporting](https://staging.review-criminal-legal-aid.service.justice.gov.uk/reporting) and verify that sticky headers remain visible while scrolling through each report. Confirm that the updated column headings remove the need for horizontal scrolling, and verify that sorting continues to work correctly. At smaller viewport widths (or when browser zoom causes the responsive breakpoint to be reached), confirm that the sticky headers are disabled as expected.

### DIFF
--- a/app/components/data_table/header_row_component.rb
+++ b/app/components/data_table/header_row_component.rb
@@ -16,19 +16,7 @@ module DataTable
     end
 
     def call
-      safe_join([
-                  caption,
-                  tag.tr(**html_attributes) { safe_join(cells) }
-                ])
-    end
-
-    def caption
-      tag.caption do
-        tag.span(
-          I18n.t('table_headings.sortable_caption'),
-          class: 'govuk-visually-hidden'
-        )
-      end
+      tag.tr(**html_attributes) { safe_join(cells) }
     end
 
     private
--- a/app/components/data_table_component.rb
+++ b/app/components/data_table_component.rb
@@ -8,4 +8,15 @@ class DataTableComponent < GovukComponent::TableComponent
   def call
     tag.table(**html_attributes) { safe_join([caption, colgroups, head, sortable_head, bodies, foot]) }
   end
+
+  private
+
+  def caption
+    tag.caption do
+      tag.span(
+        I18n.t('table_headings.sortable_caption'),
+        class: 'govuk-visually-hidden'
+      )
+    end
+  end
 end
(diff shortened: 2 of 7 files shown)


app/components/data_table/header_row_component.rb
```diff
@@ -16,19 +16,7 @@ def initialize(sorting:, filter:, classes: [])
...
-    end
-
-    def caption
-      tag.caption do
-        tag.span(
-          I18n.t('table_headings.sortable_caption'),
-          class: 'govuk-visually-hidden'
-        )
-      end
+      tag.tr(**html_attributes) { safe_join(cells) }
```
amazimbe, line 19: The removed code was syntactically incorrect. An html table should have only one caption and this code added a new caption for each row and caused browsers to misinterpret the html.

app/components/data_table_component.rb
```diff
@@ -8,4 +8,15 @@ def initialize(classes: [])
   def call
     tag.table(**html_attributes) { safe_join([caption, colgroups, head, sortable_head, bodies, foot]) }
   end
+
+  private
+
+  def caption
+    tag.caption do
+      tag.span(
+        I18n.t('table_headings.sortable_caption'),
```
amazimbe, line 17: Captions only make sense at the table level

config/locales/en/reporting.yml
```diff
@@ -59,7 +59,7 @@ en:
     application_type: Application type
     applications_closed: Number of closed applications
     applications_received: Applications received
-    assigned_to_user: "from<br>list"
+    assigned_to_user: "from list"
```
amazimbe, line 62: The space allows splitting the table header across lines when space is insufficient. Apparently that's what <br> was supposed to achieve but it ended up with the words joining up and then not splitting onto new line.

app/components/data_table/header_row_component.rb
EdwinKruglov: Well spotted!

EdwinKruglov (approved):

PR #1188, opened 2026-07-31
Upgrade Header along with justice and gov dependencies

AndyDouglasGit:
## Description of change

image

Utilise a gem for the local header to replace than the local written version. This means:

- the header will be standardised across all products (including our own)
- we will no longer have to maintain handcrafted code for that part of the app
- we can now upgrade our gov and moj dependencies along wth their dependencies

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-2315

## Notes for reviewer

There will be visual differences from the previous version:
- different logo
- gov uk removed
- name of the app
- the bottom border of the header no longer changes color with the environment

but the links available and their functionality will be the same.

### Before changes:

image

### After changes:

Note: user does have different permissions to above so different options available
image


app/views/layouts/errors.html.erb
```diff
@@ -1,7 +1,8 @@
 <% content_for?(:page_title) ? yield(:page_title) : fallback_title %>
 
 <% content_for(:header) do %>
-  <% govuk_header(container_classes: ["app-banner-#{HostEnv.env_name}"]) %>
+  <%= moj_header(organisation_name: t('service.name'), url: root_path) do |header| %>
```
EdwinKruglov: Since the block is empty, you can just do:
```ruby
<%= moj_header(organisation_name: t('service.name'), url: root_path) %>
```

Same for `app/views/layouts/external.html.erb`.

EdwinKruglov (changes requested):

EdwinKruglov (dismissed):

mooktakim-moj (dismissed):

EdwinKruglov (approved):

